PR #2, opened 2020-06-25
Batch effect correction

ben-heil:
This PR implements the logic for doing batch correction with Limma and ComBat, then looks at how their results affect the distribution of the blood data in the compendium

The part that needs to be looked over the most carefully is the batch effect correction logic. I'm pretty sure I did it correctly, but reading R documentation is hard for me

Changes:
- Added basic files required for installing the repo as a module
- Automated downloading the human compendium data from refine.bio and setting up the data directory
- Created a notebook looking at the effects of batch effect correction
- Wrote utility scripts to clean up the notebook's logic and make future analysis easier


environment.yml
```diff
@@ -0,0 +1,223 @@
...
+  - xorg-xproto=7.0.31
+  - xz=5.2.5
+  - zeromq=4.3.2
+  - zipp=3.1.0
+  - zlib=1.2.11
+  - zstd=1.4.4
+  - pip:
+    - pyqt5-sip==4.19.18
+    - pyqtchart==5.12
+    - pyqtwebengine==5.12.1
```
ajlee21, line 223: I think you are more of the conda env expert, so I'm curious. This looks like a lot of packages to define. Are most of these dependencies that don't need to be explicitly added here? Or is it better to have the yml be exhaustive?

notebook/converted/batch_effect_correction.py
```diff
@@ -0,0 +1,424 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# # Evaluation of Batch Effect Correction
+# This notebook asks whether batch effect correction works for correcting inter-study differences and/or interplatform differences.
```
ajlee21: How will you be evaluating if the correction is "working"? Perhaps this will come later..

notebook/converted/batch_effect_correction.py
```diff
@@ -0,0 +1,424 @@
...
+from plotnine import ggplot, geom_point, aes
+from tqdm.notebook import tqdm
+
+from saged import utils
+
+
+# In[2]:
+
+
+def return_unlabeled():
```
ajlee21, line 26: Perhaps this is something you will be planning to use in the future, but I guess I'm a bit confused as to why there is a function that just returns a string as opposed to just setting this string as a variable to be passed in...

saged/utils.py
```diff
@@ -0,0 +1,171 @@
...
+def parse_map_file(map_file_path: str) -> Dict[str, str]:
+    '''Create a sample: label mapping from the pickled file output by label_samples.py
+
+    Arguments
+    ---------
+    map_file_path: The path to a pickled file created by label_samples.py
+
+    Returns
+    -------
+    sample_to_label: A string to string dict mapping sample ids to their corresponding label string
```
ajlee21, line 53: NICE! Very clear

notebook/converted/batch_effect_correction.py
```diff
@@ -0,0 +1,424 @@
...
+valid_sample_ids = [id_ for id_ in sample_ids if id_ in header_ids]  
+valid_sample_ids = set(valid_sample_ids)
+print(len(valid_sample_ids))
+
+
+# ## Read in expression data
+# Now that we know which samples we're interested in, let's read those samples from the compendium.
+# This is doable with pandas, but subsetting a dataframe is extremely slow for some reason.
+# Manually writing the subsetted tsv to a file and loading it should be substantially faster and
+# less memory intensive
```
ajlee21, line 85: I might have missed something, but is there a reason you cannot use `df.loc[valid_sample_ids]`?

notebook/converted/batch_effect_correction.py
```diff
@@ -0,0 +1,424 @@
...
+# In[13]:
+
+
+embedding_df = pd.DataFrame.from_dict({'sample': list(compendium_df.index),
+                                       'pc1': list(pca_embedding[:,0]),
+                                       'pc2': list(pca_embedding[:,1]),
+                                      })
+
+
+# In[ ]:
```
ajlee21: extra cell

notebook/converted/batch_effect_correction.py
```diff
@@ -0,0 +1,424 @@
...
+ggplot(embedding_df, aes(x='study_pc1', y='study_pc2', color='label')) + geom_point()
+
+
+# In[35]:
+
+
+ggplot(embedding_df, aes(x='platform_pc1', y='platform_pc2', color='label')) + geom_point()
+
+
+# ## Limma Results
```
ajlee21, line 347: Very nice. It might be helpful to have a side by side plot for before and after correction to help with interpretation.

saged/utils.py
```diff
@@ -0,0 +1,171 @@
...
+        if tissue in BLOOD_KEYS and sample not in labeled_samples:
+            unlabeled_samples.append(sample)
+    unlabeled_samples = set(unlabeled_samples)
+
+    sample_ids = labeled_samples.union(unlabeled_samples)
+
+    return sample_ids
+
+
+def run_combat(expression_values: np.array, batches: List[str]) -> np.array:
```
ajlee21, line 128: This look correct to me. This inputs you're passing in are gene x sample and the batches are the platform or study labels for each sample in the same order as the expression data.

notebook/converted/batch_effect_correction.py
```diff
@@ -0,0 +1,424 @@
...
+
+
+study_corrected = utils.run_combat(expression_values, study_vector)
+platform_corrected = utils.run_combat(expression_values, platform_vector)
+
+
+# In[37]:
+
+
+study_corrected = study_corrected.transpose()
```
ajlee21, line 364: Very minor, but `.T` also works

ajlee21 (approved):
Very interesting. Have you tried looking to see if there are confounders between platform/study and your traits? I wonder if you're correcting out the signal because of that.

Its definitely a trade off between correcting for these technical variables and the signal...do you have a proposal for what you plan to do in this case? Happy to talk more offline about this too

environment.yml
ben-heil: Most are dependencies, so don't need to be explicitly added. Since the environment file is used programmatically I prefer to list all the versions in the environment for safety

notebook/converted/batch_effect_correction.py
ben-heil: In this case "working" is just "did the inter-study differences go away or not", but I agree that I should be more clear

notebook/converted/batch_effect_correction.py
ben-heil: `defaultdict`s take in a callable that provides a default value instead of the value itself. I could use `sample_to_label = collections.defaultdict(lambda: "unlabeled", sample_to_label)` to do the same thing, but I felt like creating a small function was clearer

saged/utils.py
ben-heil: Thanks!

notebook/converted/batch_effect_correction.py
ben-heil: The human compendium is over 100GB on disk, and would take up a little more space in memory. As a result it makes sense to write the subsetted rows directly to a file rather than getting pandas involved

saged/utils.py
ben-heil: Great!

notebook/converted/batch_effect_correction.py
ben-heil: Good idea!

PR #4, opened 2020-06-29
Base classes

ben-heil:
This PR creates the abstract base classes that define the API for models and datsets in the benchmark.

The Model classes will inherit from an API that looks like the sklearn API, and the dataset API is based on pytorch datasets. 
The datasets also fall into two categories, labeled and unlabeled, with different required functions for each.

Questions:
- Is there a clearer way of naming `subset_samples_for_label` and `subset_samples_to_labels`?
- I feel like there should be a better way of recording model results than creating my own class, but I can't think of anything that works for both sklearn and pytorch models
- The only two required functions for the `ExpressionModel` class are `fit` and `predict`. Am I missing anything that should be present in all models?

### DIFF
--- a/saged/datasets.py
+++ b/saged/datasets.py
@@ -79,17 +79,21 @@ class ExpressionDataset(ABC):
         """
         raise NotImplementedError
 
-    def subset_studies(self, fraction: float) -> float:
+    def subset_studies(self, fraction: float = None, num_studies: int = None) -> float:
         """
         This method is similar to `subset_samples`, but removes entire studies until the
         fraction of data remaining is less than the fraction passed in.
 
         As this will almost never result in the fraction being met exactly, the data fraction
-        that actually comes out of the subsetting will be returned
+        that actually comes out of the subsetting will be returned.
+
+        Either `fraction` or `num_studies` must be specified. If both are specified,
+        `num_studies` will be given preference.
 
         Arguments
         ---------
         fraction: The fraction of the samples to keep
+        num_studies: The number of studies to keep
 
         Returns
         -------
@@ -126,13 +130,13 @@ class ExpressionDataset(ABC):
     @abstractmethod
     # For more info on using a forward reference for the type, see
     # https://github.com/python/mypy/issues/3661#issuecomment-313157498
-    def get_cv_splits(self, num_splits) -> Sequence["ExpressionDataset"]:
+    def get_cv_splits(self, num_splits: int) -> Sequence["ExpressionDataset"]:
         """
         Split the dataset into a list of smaller dataset objects with a roughly equal
         number of samples in each.
 
-        If multiple studies are present in the dataset, each dataset should only
-        have data from a single study
+        If multiple studies are present in the dataset, each study should only
+        be present in a single fold
 
         Arguments
         ---------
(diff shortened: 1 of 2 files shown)


saged/models.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    A structure for storing the metrics corresponding to a model's training.
+    The ModelResults class requires a loss and another metric such as accuracy or F1 score.
+    Other metrics can be tracked via the `add_other` method
+    """
+    def __init__(self,
+                 model_name: str,
+                 progress_type: str,
+                 loss_type: str,
+                 accuracy_metric_type: str
```
jjc2718: Trying to think of how an accuracy metric would generalize to regression, since accuracy isn't really a thing if you're predicting real values rather than discrete classes.

Do you think you'll only be using this evaluation pipeline to benchmark classification? If you plan to benchmark regression as well, maybe it's best to just add metrics other than the loss using `add_other` (which for classification could be accuracy or F1 score, and for regression could be R^2, etc). 

Just a thought though, I don't feel too strongly about this. I'm mostly focused on classification in my current project, and I think you are too so maybe focusing on classification is fine for now.

saged/models.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __init__(self) -> None:
+        """
+        Standard model init function. We use pass instead of raising a NotImplementedError
+        here in case inheriting classes decide to call `super()`
+        """
+        pass
+
+    @abstractmethod
+    def fit(self, dataset: LabeledDataset) -> ModelResults:
```
jjc2718, line 110: I think just having `fit` and `predict` methods here is good. For some models you'll want to implement additional methods like `decision_function`, `predict_proba`, etc, but not all models have those so keeping this class as general as possible seems like the way to go.

One thing you could consider is having `get_params` and `set_params` functions, which would allow saving (for instance) the weights of a trained NN or linear model and applying them to data later. But I'm not sure how this would work for something like k-nearest neighbors which doesn't really fit parameters, so maybe it's best to just add these in the inheriting classes.

saged/datasets.py
```diff
@@ -0,0 +1,209 @@
...
+from typing import Sequence, Tuple, List
+
+import numpy as np
+
+
+class ExpressionDataset(ABC):
+    """
+    The base dataset defining the API for datasets in this project. This class is
+    based on the Dataset object from pytorch
+    (https://pytorch.org/docs/stable/_modules/torch/utils/data/dataset.html#Dataset
```
jjc2718:
```suggestion
    (https://pytorch.org/docs/stable/_modules/torch/utils/data/dataset.html#Dataset)
```

saged/datasets.py
```diff
@@ -0,0 +1,209 @@
...
+
+        Returns
+        -------
+        X: The gene expression data in a genes x samples array
+        y: The label corresponding to each sample in X
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def subset_samples(self, fraction: float) -> None:
```
jjc2718: Maybe these subset functions should take a random seed as an argument? Unless you want to leave the responsibility of shuffling the data to the function caller and then just take the top fraction every time, but then that should probably be documented somewhere.

saged/datasets.py
```diff
@@ -0,0 +1,209 @@
...
+        """
+        Limit the amount of data available to be returned in proportion to
+        the total amount of data that was present in the dataset.
+        That is to say that calling subset_data_by_fraction twice will not cause the Dataset
+        object to have multiplicatively less data, it will have the fraction of data specified
+        in the second call
+
+        Arguments
+        ---------
+        fraction: The fraction of the samples to keep
```
jjc2718, line 77: Are you assuming here that the function will modify the data structure that's passed in, rather than creating a copy and modifying and returning that? Or is that up to the inheriting classes?

If you're making an assumption about this, I'd suggest making it clear in the function documentation (same for the `subset_studies` function).

saged/datasets.py
```diff
@@ -0,0 +1,209 @@
...
+        -------
+        train: The dataset with around the amount of data specified by train_fraction
+        test: The dataset with the remaining data
+        """
+        raise NotImplementedError
+
+
+class LabeledDataset(ExpressionDataset):
+    @abstractmethod
+    def subset_samples_for_label(self, fraction: float, label: str) -> None:
```
jjc2718: I think these names are fine, they make sense to me. Maybe you could be slightly more explicit with something like `subset_samples_within_single_label` and `subset_all_samples_to_labels`? I don't think those names are a huge improvement (if any) over what you have, though.

jjc2718 (approved):
This looks like a good start to me! I think this will probably evolve as we get a better idea of the use cases for these classes, but you have everything I can think of here as a starting point.

Comments are mostly just things to think about, generally I'm fine with what you have.

saged/models.py
ben-heil: I'll probably switch it back to `metric_type`. I feel like just using `metric` as a descriptor is vague, but iirc that's how Tensorflow does it.

saged/datasets.py
ben-heil: Good catch

saged/datasets.py
ben-heil: I think it makes sense to leave it to the inheriting classes. The function should probably return the Dataset object afterwards either way to allow chaining calls, so I'll add that

saged/datasets.py
ben-heil: I think I'll stick with the current versions just because they're shorter, I'll continue thinking on it though

PR #5, opened 2020-07-08
Refinebio labeled dataset

ben-heil:
This PR has three main parts:
1. It implements the RefineBioLabeledDataset class in dataset.py
2. It creates testing infrastructure and tests the class
3. It generates data to be used in the tests

It also adds some code to `utils.py` to make implementation more straightforward


saged/generate_test_data.py
```diff
@@ -0,0 +1,78 @@
+""" This file generates a dataset to be used in testing """
```
jjc2718, line 1: It looks like in order for the testing to work, this script has to be run once. Is that right? If so, do you want to document this in the README (or somewhere else) so new users can run it when they set up the repo?

saged/generate_test_data.py
```diff
@@ -0,0 +1,78 @@
...
+
+random.seed(42)
+np.random.seed(42)
+
+NUM_SAMPLES = 200
+NUM_GENES = 5
+
+current_dir = os.path.dirname(os.path.abspath(__file__))
+
+compendium_path = os.path.join(current_dir, '../data/HOMO_SAPIENS.tsv')
```
jjc2718, line 19: Will this directory always exist?

saged/generate_test_data.py
```diff
@@ -0,0 +1,78 @@
...
+
+study_assignments = random.choices(study_list, weights=study_probs, k=NUM_SAMPLES)
+
+# Generate formatted metadata mapping studies to samples
+fake_metadata = {'experiments': {study: {'sample_accession_codes': []} for study in study_list}}
+for sample, study in zip(samples, study_assignments):
+    fake_metadata['experiments'][study]['sample_accession_codes'].append(sample)
+
+# Write formatted metadata to a file
+out_directory = os.path.join(current_dir, '../test/data')
```
jjc2718, line 41: Same question as above, will the `../test` and `../test/data` directories always exist here?

saged/utils.py
```diff
@@ -169,3 +172,111 @@ def run_limma(expression_values: np.array,
...
+    sample_to_label: A string to string dict mapping sample ids to their
+        corresponding label string. E.g. {'GSM297791': 'sepsis'}
+    """
+    sample_to_label = {}
+    label_to_sample = None
+    with open(label_file_path, 'rb') as map_file:
+        label_to_sample, _ = pickle.load(map_file)
+
+    for label in label_to_sample:
+        for sample in label_to_sample[label]:
```
jjc2718, line 196: Are you sure each sample will have exactly one label? Even if that is the case, it might be good to put an `assert` or other type of error checking code here before you put it into the dict. Otherwise, whatever was in the dict for the sample before will get overwritten silently.

saged/utils.py
```diff
@@ -169,3 +172,111 @@ def run_limma(expression_values: np.array,
...
+    sample_to_study:
+        The mapping from sample accessions to the study they are a member of
+    """
+    experiments = metadata_json['experiments']
+    id_set = set(sample_ids)
+
+    sample_to_study = {}
+    for study in experiments:
+        for accession in experiments[study]['sample_accession_codes']:
+            if accession in id_set:
```
jjc2718, line 260: Will this ever not be the case? (Stated otherwise, is this function used to filter to certain samples/accession codes, or is this just to be cautious?)

saged/datasets.py
```diff
@@ -231,3 +244,443 @@ def get_all_data(self) -> np.array:
...
+        sample_to_label: A mapping between sample accessions and their phenotype labels
+        sample_to_study: A mapping between sample accessions and their study accessions
+        """
+        self.all_expression = expression_df
+        self.sample_to_label = sample_to_label
+        self.sample_to_study = sample_to_study
+
+        # We will handle subsetting by creating views of the full dataset.
+        # Most functions will access current_expression instead of all_expression.
+        self.current_expression = expression_df
```
jjc2718, line 274: Do you want to set `self.data_changed = False` here? Or does it not matter?

saged/datasets.py
```diff
@@ -231,3 +244,443 @@ def get_all_data(self) -> np.array:
...
+
+        Returns
+        -------
+        self: The subsetted version of the dataset
+        """
+        self.data_changed = True
+        random.seed(seed)
+
+        all_samples = self.get_samples()
+        samples_without_label = [sample for sample in all_samples
```
jjc2718: If you want, you could probably partition `all_samples` in one pass (see, for example: https://stackoverflow.com/a/8793838).

That might be slightly less readable/understandable though (and not immediately clear to me whether it would actually be faster), so it might be best to stick with what you have if you don't expect this to be slow.

jjc2718 (approved):
Looks good to me! Just a few small comments, nothing major.

saged/datasets.py
jjc2718: Or something super simple like this is also an option: https://stackoverflow.com/a/58029106

saged/generate_test_data.py
ben-heil: The testing data files are tracked by git, so the users don't need to do anything for the tests to run correctly. I'm tracking the file so people can see how the data was generated and tweak it if they are inclined to

saged/generate_test_data.py
ben-heil: They will exist by default, so as long as the user doesn't delete them the relative path should work

saged/utils.py
ben-heil: Yes, the metadata tends to correspond to the entire human compendium while I'm only interested about a subset at a time. I could write code to manually create subset files of the metadata, but it's easier to handle it here.

The mapping would work either way, but this makes it an order of magnitude smaller

saged/utils.py
ben-heil: The current version of the labeling script I used only allows one label per sample. That being said, it could change in the future so you're right that I should add an assert statement to make sure that is true

saged/datasets.py
ben-heil: If `self.data_changed` were set to False in the init function, it would throw an error when we try to `get_studies()` before something else sets self.data_chaged to True. I could set `self.data_chaged = True` in the init function, but that seems a little confusing. I'll stick with checking whether the attribute `data_changed` is present, but I'll add a comment explaining why that's necessary

saged/datasets.py
ben-heil: Ooh good idea

PR #6, opened 2020-07-13
Implemented and tested refinebio unlabeled dataset

ben-heil:
This PR creates the unlabeled equivalent of the labeled dataset from the last PR, and adds some tests to ensure randomization is working correctly in the datasets.

For the most part the PR looks bigger than it is. I realized that it made sense to have the labeled dataset inherit from the unlabeled version, so all the functions that work for both objects got moved over to the new class. 

I think I have abstracted out most of the unnecessary code repetition between the classes, but if I missed something let me know.

### DIFF
--- a/saged/datasets.py
+++ b/saged/datasets.py
@@ -476,8 +476,48 @@ class RefineBioUnlabeledDataset(UnlabeledDataset):
         self.data_changed = True
         return self
 
-    # For more info on using a forward reference for the type, see
-    # https://github.com/python/mypy/issues/3661#issuecomment-313157498
+    def get_cv_expression(self, num_splits: int, seed: int = 42) -> List[pd.DataFrame]:
+        """
+        Split the expression present in the dataset into `num_splits` partitions for
+        use in cross-validation
+
+        Arguments
+        ---------
+        num_splits: The number of groups to split the dataset into
+        seed: The seed for the random number generator involved in subsetting
+
+        Returns
+        -------
+        cv_dataframes: The expression dataframe views for each cv split
+        """
+        random.seed(seed)
+
+        samples = self.get_samples()
+        studies = self.get_studies()
+        shuffled_studies = random.sample(studies, len(studies))
+
+        base_study_count = len(studies) // num_splits
+        leftover = len(studies) % num_splits
+        study_index = 0
+
+        cv_dataframes = []
+        for i in range(num_splits):
+            study_count = base_study_count
+            if i < leftover:
+                study_count += 1
+
+            current_studies = shuffled_studies[study_index:study_index+study_count]
+            study_index += study_count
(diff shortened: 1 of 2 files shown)


saged/datasets.py
```diff
@@ -477,15 +476,10 @@ def subset_studies(self,
         self.data_changed = True
         return self
 
-    # For more info on using a forward reference for the type, see
-    # https://github.com/python/mypy/issues/3661#issuecomment-313157498
-    def get_cv_splits(self, num_splits: int, seed: int = 42) -> Sequence["ExpressionDataset"]:
+    def get_cv_expression(self, num_splits: int, seed: int = 42) -> List[pd.DataFrame]:
```
jjc2718, line 479: Do you expect this function to be used externally, or does it mostly exist as a helper function for `get_cv_splits`? 

You could make it a private method if you don't want it to be used externally (i.e. just name it with a leading underscore, since Python doesn't really have private methods per se: https://stackoverflow.com/a/6930223). But if you think there's a case for using it sometimes instead of `get_cv_splits` then what you have is fine.

test/test_datasets.py
```diff
@@ -102,29 +131,43 @@ def test_get_studies():
...
 
-            subset_dataset = dataset.subset_studies(fraction, num_studies, seed)
+            subset_dataset = dataset.subset_studies(fraction, num_studies, seed=1)
             samples = subset_dataset.get_samples()
 
             assert len(samples) > len(all_samples) * fraction
 
         dataset.reset_filters()
 
+        # There are only six studies, so tests will spuriously fail when selecting many studies
```
jjc2718, line 159: Couldn't this still fail even when there's (for example) one study, if the one you select below happens to be the same as the one you selected before? Or am I missing something?

I guess if you're using the same seeds (1 and 2) every time and it doesn't fail the first time it won't fail in the future, so maybe this is fine. It seems like it could fail if you changed the seeds, though.

jjc2718 (approved):
Nice work! Two small comments, but otherwise looks good to me.

saged/datasets.py
ben-heil: That's a good point, I'll switch it to a private method

test/test_datasets.py
ben-heil: It could still fail, but it doesn't with the given seeds. The second condition (the two subsets from the same seed should be the same) should always hold, so I guess it didn't strictly need to be in the if statement. On the other hand the first condition indicates success if it is ever true, because there has to be something nondeterministic going on to get different results. 

Ideally I'd like to just run the code fifty times with fifty different seeds, but there doesn't seem to be a built-in way to do that and I'm satisfied enough with it working now to not deal with the loop overhead

PR #7, opened 2020-07-21
Create Supervised Model

ben-heil:
This PR creates and tests the `PytorchSupervised` class. It turns out there are a lot of things required to train a neural network, so there are modifications in a number of places.

Changes:
- Tell git to ignore a secrets.yml file that stores things like API tokens
- Add instructions for installing neptune to the readme
- Update the environment file with new pytorch and neptune dependencies
- Add encoding logic to the `RefineBioLabeledExpression` class to allow training and adjust classes to return in samples x genes format
- Implement `PytorchSupervised` and its underlying model `ThreeLayerClassifier`
- Create a configuration file to keep track of how test models should be trained
- Write model tests
- Write util tests

### DIFF
--- a/saged/datasets.py
+++ b/saged/datasets.py
@@ -6,6 +6,7 @@ from typing import Sequence, Tuple, List, Union, Set, Dict
 
 import numpy as np
 import pandas as pd
+from sklearn import preprocessing
 
 from saged import utils
 
@@ -694,6 +695,11 @@ class RefineBioLabeledDataset(RefineBioUnlabeledDataset):
         self.sample_to_label = sample_to_label
         self.sample_to_study = sample_to_study
 
+        encoder = preprocessing.LabelEncoder()
+        labels = [sample_to_label[sample] for sample in expression_df.columns]
+        encoder.fit(labels)
+        self.encoder = encoder
+
         self.current_expression = expression_df
 
     @classmethod
@@ -751,8 +757,10 @@ class RefineBioLabeledDataset(RefineBioUnlabeledDataset):
         """
         sample_id = self.get_samples()[idx]
         sample = self.current_expression[sample_id].values
-        label = np.array(self.sample_to_label[sample_id])
-        return sample, label
+        label = self.sample_to_label[sample_id]
+        encoded_label = self.encoder.transform([label])  # Braces are necessary to be a 1d array
+
+        return sample, encoded_label
 
     def get_all_data(self) -> Tuple[np.array, np.array]:
         """
@@ -768,9 +776,17 @@ class RefineBioLabeledDataset(RefineBioUnlabeledDataset):
         X = self.current_expression.values.T
         sample_ids = self.get_samples()
         labels = [self.sample_to_label[sample] for sample in sample_ids]
-        y = np.array(labels)
+        y = self.encoder.transform(labels)
 
         return X, y
+
+    def recode(self) -> None:
+        """
+        Retrain the encoder to contain only the labels present in current_expression instead
+        of all the labels in the dataset
+        """
+        labels = [self.sample_to_label[sample] for sample in self.current_expression.columns]
+        self.encoder.fit(labels)
 
     def map_labels_to_counts(self) -> Dict[str, int]:
         """
@@ -784,7 +800,9 @@ class RefineBioLabeledDataset(RefineBioUnlabeledDataset):
 
         sample_ids = self.get_samples()
         labels = np.array([self.sample_to_label[sample] for sample in sample_ids])
+
         unique_elements, element_counts = np.unique(labels, return_counts=True)
+
         for label, count in zip(unique_elements, element_counts):
             label_counts[label] = count
 
--- a/saged/datasets.py
+++ b/saged/datasets.py
@@ -695,10 +695,10 @@ class RefineBioLabeledDataset(RefineBioUnlabeledDataset):
         self.sample_to_label = sample_to_label
         self.sample_to_study = sample_to_study
 
-        encoder = preprocessing.LabelEncoder()
+        label_encoder = preprocessing.LabelEncoder()
         labels = [sample_to_label[sample] for sample in expression_df.columns]
-        encoder.fit(labels)
-        self.encoder = encoder
+        label_encoder.fit(labels)
+        self.label_encoder = label_encoder
 
         self.current_expression = expression_df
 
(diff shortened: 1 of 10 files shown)


saged/datasets.py
```diff
@@ -694,6 +695,11 @@ def __init__(self,
         self.sample_to_label = sample_to_label
         self.sample_to_study = sample_to_study
 
+        encoder = preprocessing.LabelEncoder()
```
jjc2718: Calling this `encoder` may get a little confusing if you're planning to generate data using an autoencoder in the future. Maybe `label_encoder` would be clearer?

saged/datasets.py
```diff
@@ -342,9 +343,9 @@ def get_all_data(self) -> Tuple[np.array, np.array]:
 
         Returns
         -------
-        X: The gene expression data in a genes x samples array
+        X: The gene expression data in a samples x genes array
```
jjc2718, line 346: Does the data from refine.bio and other sources come in genes x samples format? I can see it being a bit confusing that you're storing it internally as a genes x samples dataframe, then transposing it here before returning it.

Might be more straightforward to just store it internally as a samples x genes dataframe (and require that orientation as input to your constructor). That's just my opinion though, ultimately either will work fine so I'll leave it up to you.

If you really wanted to get into the weeds of this, I'm pretty sure Pandas dataframes are stored in column-major order by default ([see here for a bit of background](https://en.wikipedia.org/wiki/Row-_and_column-major_order) and [here for details about Pandas](https://stackoverflow.com/questions/49420272/how-to-make-pandas-dataframe-fortran-type-ordered), Fortran ordered == column-major), although it's surprisingly hard to find official Pandas documentation of this. This means it would probably be faster to have the dimension you plan to slice by most often as columns, and the other as rows. It seems to me like you're planning to slice by samples more often than genes, so that would support the orientation you have now (genes x samples)...but I have no idea how much this distinction will actually matter in practice.

saged/models.py
```diff
@@ -134,3 +161,348 @@ def predict(self, dataset: np.array) -> np.array:
...
+    A wrapper class implementing the ExpressionModel API while remaining modular enough
+    to accept any supervised classifier implementing the nn.Module API
+    """
+    def __init__(self,
+                 config: dict,
+                 optimizer: torch.optim.Optimizer,
+                 loss_class: torch.nn.modules.loss._WeightedLoss,
+                 model_class: nn.Module) -> None:
+        """
+        Standard model init function. We use pass instead of raising a NotImplementedError
```
jjc2718: Are you actually using `pass` anywhere in this class? Or is this just here for future reference?

saged/models.py
```diff
@@ -134,3 +161,348 @@ def predict(self, dataset: np.array) -> np.array:
...
+
+    @classmethod
+    def load_model(classobject,
+                   checkpoint_path: str,
+                   config: dict,
+                   optimizer: torch.optim.Optimizer,
+                   loss_fn: torch.nn.modules.loss._WeightedLoss,
+                   model_class: nn.Module) -> "PytorchSupervised":
+        """
+        Read a pickeled model from a file and return it
```
```suggestion
        Read a pickled model from a file and return it
```

saged/utils.py
```diff
@@ -281,3 +285,60 @@ def get_samples_in_studies(samples: List[str],
...
+    Returns
+    -------
+    The integer labels predicted by the model
+    """
+    return torch.argmax(model_output, dim=-1)
+
+
+def count_correct(outputs: torch.Tensor, labels: torch.Tensor) -> int:
+    """
+    Calculate the nubmer of correct predictions in the given batch
```
jjc2718:
```suggestion
    Calculate the number of correct predictions in the given batch
```

saged/models.py
```diff
@@ -134,3 +161,348 @@ def predict(self, dataset: np.array) -> np.array:
...
+        return x
+
+
+class PytorchSupervised(ExpressionModel):
+    """
+    A wrapper class implementing the ExpressionModel API while remaining modular enough
+    to accept any supervised classifier implementing the nn.Module API
+    """
+    def __init__(self,
+                 config: dict,
```
jjc2718, line 284: Do you know of a good way to document what needs to be in this config, and (if anything) what parameters are optional? I guess as it stands your code will throw a `KeyError` if (for example) the learning rate or seed is missing, but it might be good to have a list of what's necessary somewhere centralized.

Even just documentation pointing to the example config in `test/data/test_config.yml` might serve this purpose.

saged/models.py
```diff
@@ -134,3 +161,348 @@ def predict(self, dataset: np.array) -> np.array:
...
+        self.model = model_class(model_config)
+        lr = float(model_config['lr'])
+        weight_decay = float(model_config.get('weight_decay', 0))
+        self.optimizer = optimizer(self.model.parameters(),
+                                   lr=lr,
+                                   weight_decay=weight_decay)
+
+        self.loss_class = loss_class
+
+        self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
jjc2718: Do you always want to run PyTorch models on the GPU if one is available, or is there ever a case where you might want to override this and run things on the CPU?

I know in the past I've noticed shallow models (e.g. linear/logistic regression) in PyTorch can at times run faster on the CPU, but if you're only planning on using this class for multi-layer NNs (which is what it looks like here) you might be fine.

jjc2718 (approved):
Looks good to me! Just a few things to think about.

saged/datasets.py
ben-heil: Good call

saged/datasets.py
ben-heil: That's definitely a good point. I currently have the internal representation stored as genes x samples because that's how they're stored in the compendium file. Since my most common internal operations are subsetting by sample that should be more efficient. Both sklearn and pytorch expect their inputs to be samples x features though, so I take the transpose when returning results to keep the data consistent with what the models are expecting

saged/models.py
ben-heil: Nope, good catch. That's a holdover from the ExpressionModel abstract class that does use `pass` for its init function

saged/models.py
ben-heil: That's a good question. I think it will make sense to write a custom parser function that validates the file, but I'll wait for a bit before doing so to let the config format solidify.

I think for now I'll add a reference in the README and add a TODO to the code

saged/models.py
ben-heil: That's a good point, it makes sense to put that in the config file too

PR #8, opened 2020-07-29
Implement and Test a PCA Class

ben-heil:
This PR creates an UnsupervisedModel base class and a PCA class that inherits from it. It also tests the PCA class. Thankfully this PR is much less complex than the past couple have been.

Changes:
- Create UnsupervisedModel and PCA classes in `models.py`
- Create an `unlabeled_dataset` fixture in `test_models.py` to clean up some functions
- Write tests for PCA class

### DIFF
--- a/saged/models.py
+++ b/saged/models.py
@@ -616,6 +616,7 @@ class PCA(UnsupervisedModel):
         self.model = sklearn.decomposition.PCA(n_components=n_components,
                                                random_state=seed)
 
+    @classmethod
     def load_model(classobject, model_path):
         """
         Read a pickeled model from a file and return it
--- a/saged/models.py
+++ b/saged/models.py
@@ -506,3 +506,90 @@ class PytorchSupervised(ExpressionModel):
     def load_parameters(self, parameters: dict) -> "PytorchSupervised":
         self.model.load_state_dict(parameters)
         return self
+
+
+class UnsupervisedModel():
+    """
+    A model API defining the behavior of unsupervised models. Largely follows the sklearn model api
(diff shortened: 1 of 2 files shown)


saged/models.py
```diff
@@ -506,3 +508,180 @@ def get_parameters(self) -> Iterable[torch.Tensor]:
...
+        ---------
+        out_path: The path to the file to write the classifier to
+
+        Raises
+        ------
+        FileNotFoundError if out_path isn't openable
+        """
+        raise NotImplementedError
+
+    def fit_transform(self, dataset: UnlabeledDataset) -> UnlabeledDataset:
```
jjc2718, line 584: I think in the case of PCA this might be slower than just calling `model.fit_transform`, since your implementation has to call `get_all_data` twice which transposes the data matrix twice. I could be wrong, though - I don't have a good idea of how slow it is to take a transpose in Numpy (or Pandas).

jjc2718 (approved):
Looks good! Just one comment on your PCA class implementation.

saged/models.py
ben-heil: That's a good point, I didn't think about that. Apparently [it's a constant time operation](https://stackoverflow.com/questions/58279082/time-complexity-of-numpy-transpose). A point of interesting numpy trivia that I learned: the transpose operation is done by changing the way that functions access the underlying memory, so it doesn't actually modify the matrix at all

jjc2718: Interesting! I didn't know that. In that case what you have should be fine.

Issue #3, opened 2020-06-26
Benchmarking Code


ben-heil:
The code required to compare different models will have a lot of different components, so I'll list the elements here and check them off as they're implemented.

- [x] Create a base Dataset class
- [x] Implement a Dataset class for refinebio data
- [x] Implement extracting data for a given label
- [x] Implement subsetting data based on label
- [x] Implement train/test split and cross-validation, by study instead of by sample

- [x] Create a base Model class (probably using the sklearn API?)
- [x] Implement Models for skl methods
- [ ] Implement Models for some semi-supervised methods
- [ ] Implement non-semi-supervised versions of the same models, or separate unsupervised and supervised steps

- [ ] Implement data simulation/augmentation with VAEs
- [x] Determine a way to parallelize the benchmarking process
- [x] Write the logic to take a dataset and benchmark models on it. Probably build this into the dataset object itself?

ben-heil:
Assorted notes from planning are below.

Benchmark structure:
- One config file that controls run details. Which batch effect correction to use, which datasets to use, which models
- The dataset and models sections contain lists of config file paths
- Each dataset has a config file denoting paths to the files they need and which function to use to initialize them
- Each model has a config file listing its model class (for finding it in models.py) and its hyperparameter values to use
- Each analysis (e.g. performance of models on all classes, performance of models on single class) are their own python script
- The logic for running everything is strung together with snakemake
- Results formats are still up in the air. Probably tsv, but we need to have lists of study ids, models, etc

 ```
# Across all diseases, do we perform better or worse with embeddings                                                                                                                           
Load dataset                                                                                                                                                                                  
Split compendium data into labeled and unlabeled data                                              
Split labeled data via 5FCV                                                                        
                                                                                                    
Get the number of labels and the number of genes.                                                  
Load the supervised model from the config passed in
If an unsupervised model is called for, load the unsupervised model and embed data
Iterate over CV splits:
    train model
    evaluate model
    save results
                                                                                                                                                   
# How are we storing results?                                                                       
# How do we implement early stopping and resuming? Snakemake?                                       
```

PR #10, opened 2020-08-07
All labels benchmark

ben-heil:
This PR contains the new design for the benchmark as laid out [here](https://github.com/greenelab/saged/issues/3#issuecomment-666509142). Namely, lots of config files for controlling how things run, and general benchmark scripts that are agnostic about what type of model/dataset they are running. Unfortunately this led to a lot of code changes. I'm sorry for the size of the PR.

Changes:
- Created lots of config files
- Wrote the first benchmark, `all_label_comparison`, that trains a model to do a multiclass prediction on all labels in the dataset
- Made the infrastructure for datasets containing both labeled and unlabeled data
- Updated the `models.py` and `datasets.py` classes to take config files information
- Added more test cases to test the new functionality


saged/all_label_comparison.py
```diff
@@ -0,0 +1,126 @@
...
+                        help='The yaml formatted model configuration file for an unsupervised '
+                             'model. If omitted, the benchmark will use the original features '
+                             'for training. For more information about this file read the '
+                             'comments in the example_model.yml file',
+                        default=None)
+    parser.add_argument('--neptune_config',
+                        help='A yaml formatted file containing init information for '
+                             'neptune logging')
+    parser.add_argument('--seed',
+                        help='The random seed to be used in the experiment',
```
jjc2718: Might be good to note that this seed only controls the way the data is split, not the model initialization etc. (looks like the seed for the Pytorch model is set separately in the relevant model config).

saged/all_label_comparison.py
```diff
@@ -0,0 +1,126 @@
...
+    print('Loading all data')
+    all_data = MixedDatasetClass.from_config(**dataset_config)
+    print('Subsetting labeled data')
+    labeled_data = all_data.get_labeled()
+    print('Subsetting unlabeled data')
+    unlabeled_data = all_data.get_unlabeled()
+    print('Splitting data')
+
+    # Get fivefold cross-validation splits
+    labeled_splits = labeled_data.get_cv_splits(num_splits=5, seed=args.seed)
```
jjc2718: You could make the number of splits a command-line parameter as well, or define it in a config file somewhere. I doubt you'll need to change it much, if at all, but it probably wouldn't hurt to make it more accessible.

saged/datasets.py
```diff
@@ -320,34 +366,20 @@ def __len__(self) -> int:
...
-        -------
-        sample: The gene expression data for the given index in a genes x 1 array
-        """
-        sample_id = self.get_samples()[idx]
-        sample = self.current_expression[sample_id].values
-        return sample
-
-    def get_all_data(self) -> Tuple[np.array, np.array]:
+        new_data: The variable x samples array to overwrite the old data with. The first dimension
+                  can be anything, but the sample dimension must remain the same as in the old data
```
jjc2718: It's not just that the dimension has to be the same, but the columns have to be in the same order too, right?

saged/datasets.py
```diff
@@ -731,16 +943,59 @@ def from_paths(class_object,
...
+                # https://stackoverflow.com/questions/19125091/pandas-merge-how-to-avoid-duplicating-columns
+                cols_to_use = current_dataframe.columns.difference(combined_dataframe.columns)
+
+                # Combine dataframes
+                combined_dataframe = pd.merge(combined_dataframe,
+                                              current_dataframe[cols_to_use],
+                                              left_index=True,
+                                              right_index=True,
+                                              how='outer')
+                # Combine labels (double start expands a dictionary into key-value pairs)
```
jjc2718:
```suggestion
                # Combine labels (double star expands a dictionary into key-value pairs)
```
Also, do you want this comment in the function in `RefineBioUnlabeledDataset` too?

saged/datasets.py
```diff
@@ -952,3 +1223,266 @@ def train_test_split(self,
...
```
jjc2718:
```suggestion
                # Combine labels (double star expands a dictionary into key-value pairs)
```

saged/models.py
```diff
@@ -185,11 +201,16 @@ class LogisticRegression(ExpressionModel):
...
+                 seed: int,
+                 **kwargs,
+                 ) -> None:
         """
         The initializer the LogisticRegression class
 
-        seed: The random seed to be used by the model
+        Arguments
+        ---------
+        seed: The random seed ot use in training
```
```suggestion
        seed: The random seed to use in training
```

saged/models.py
```diff
@@ -185,11 +201,16 @@ class LogisticRegression(ExpressionModel):
     base acceptable functions for models in this module's benchmarking code
     """
 
-    def __init__(self, seed: int = 42) -> None:
+    def __init__(self,
+                 seed: int,
+                 **kwargs,
+                 ) -> None:
         """
         The initializer the LogisticRegression class
```
```suggestion
        The initializer of the LogisticRegression class
```

saged/models.py
```diff
@@ -410,8 +501,6 @@ def fit(self, dataset: LabeledDataset) -> "PytorchSupervised":
             self.loss_fn = self.loss_class(weight=None)
 
         if log_progress:
-            # Set up the neptune experiment
-            utils.initialize_neptune(self.config)
```
jjc2718, line 414: Is this function still being used anywhere?

test/test_datasets.py
```diff
@@ -344,3 +386,95 @@ def test_subset_samples_to_labels(labeled_datasets, labels):
...
+
+def test_get_samples(all_datasets):
+    # Get sample names
+    current_dir = os.path.dirname(os.path.abspath(__file__))
+    expression_path = os.path.join(current_dir, 'data/test_expression.tsv')
+    with open(expression_path) as expression_file:
+        true_samples = set(expression_file.readline().strip().split('\t'))
+
+    for dataset in all_datasets:
+        print(type(dataset))
```
jjc2718: Do you want to leave this print statement in?

jjc2718 (approved):
Looks good! My comments are mainly just typo fixes and small questions.

I assume you haven't actually run the benchmark script yet (so this is just a proof of concept/prototype of what you'll be running)? Or have you?

saged/all_label_comparison.py
ben-heil: Good point

saged/all_label_comparison.py
ben-heil: Will do

saged/datasets.py
ben-heil: That's true, updated the comment

saged/datasets.py
ben-heil: Went ahead and added it to `RefineBioUnlabeledDataset`

saged/models.py
ben-heil: Eek, sorry for all the typos

saged/models.py
ben-heil: Yeah, I moved it to `all_labels_comparison`, with the idea that each benchmark will be responsible for setting up its own connection with neptune. It also makes more it so that it's unnecessary to pass all the neptune configuration information to every model

test/test_datasets.py
ben-heil: Good catch, went ahead and removed the debug print statements

PR #11, opened 2020-08-13
Fix memory leak in PytorchSupervised

### LINKED ISSUE #9: PytorchSupervised Memory Leak
https://ui.neptune.ai/ben-heil/saged/e/SAG-4/monitoring and https://ui.neptune.ai/ben-heil/saged/e/SAG-5/monitoring correspond to the first and second fold of cross validation in `all_label_comparison.py`. The second fold starts with a higher GPU memory usage, and as a result errors out. 

This implies there is gpu memory being held across folds. It isn't immediately apparent why this is happening, and it doesn't fit in the scope of the current PR, so I've opened this issue.

ben-heil:
It turns out that making pytorch models members of a larger class leads to references between the optimizer and model parameters not being deleted or something. Long story short, it's necessary to manually delete the model and optimizer to keep the old tensors from haunting you/causing memory leaks. 

If only there were a function that acted like the opposite of an initializer called `__del__` or something. Unfortunately I tried that and deleting the model and optimizer from `__del__` doesn't work while writing a custom function to do the same thing does *shrug*. I would expect it to be something to do with variable scope, but even if you explicitly call `del` on the model the tensors live on.

This solution works, though the fact that writing a `__del__` function didn't is confusing to me.

Closes #9

### DIFF
--- a/saged/all_label_comparison.py
+++ b/saged/all_label_comparison.py
@@ -110,8 +110,9 @@ if __name__ == '__main__':
 
         # Train the model on the training data
         supervised_model.fit(train_data)
-
         predictions, true_labels = supervised_model.evaluate(val_data)
+
+        supervised_model.free_memory()
 
         # TODO more measures than Top-1 accuracy
         accuracy = sklearn.metrics.accuracy_score(predictions, true_labels)
--- a/saged/models.py
+++ b/saged/models.py
@@ -134,6 +134,15 @@ class ExpressionModel(ABC):
         """
         raise NotImplementedError
 
+    def free_memory(self) -> None:
+        """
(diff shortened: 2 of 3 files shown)


jjc2718 (approved):
Looks good! Nice work tracking this down.

I'll have to keep this in mind for the future. I'm with you, not sure why adding this to the `__del__` function didn't work.

ben-heil:
> Looks good! Nice work tracking this down.
> 
> I'll have to keep this in mind for the future. I'm with you, not sure why adding this to the `__del__` function didn't work.

Thanks! I have no idea why it didn't work either. I made sure it was getting called, and it was called at the same time as `free_memory`. I probably just don't understand python's garbage collector well enough

PR #12, opened 2020-08-14
Make Transform Act on a Copy

ben-heil:
The `transform` method of unsupervised models previously modified the original dataset instead of a copy. As a result, the folds would be modified during CV, and the merged datasets on the second fold would contain both dimensionally reduced and raw data.

Now that `transform` acts on copies this doesn't happen anymore.


jjc2718 (approved):
Looks good to me!

PR #13, opened 2020-08-18
Snakemake

### LINKED ISSUE #14: `utils.load_compendium_data` is slow
For small models ~40 percent of the runtime (~3.5 minutes) is spend just loading the subset compendium into a dataframe. This appears to be due to the slowness of the `pandas.read_csv` function as opposed to a disk reading limitation, because loading the file into memory with `compendium_file.readlines()` only takes 8 seconds. 

Potential fixes:
- [ ] Pickle the dataframe and load it
- [ ] Load the file into a numpy array and use the numpy constructor for the dataframe
- [ ] Try dask again for parallelism since we appear to be CPU bound instead of disk bound

ben-heil:
Sorry for the frequent pull requests, but the current project portion lends itself to small self-contained features.

This PR writes a Snakefile which uses snakemake to run all the items in the benchmark (just `all_labels_comparison.py` for now). It also speeds up loading the main dataset because a ton of time was sunk into loading it on each run (closes #14).

If you're new to snakemake, everything is backwards of how you would expect. The `all` rule lists all the files you want to create, then the subsequent rules define how to make them. Snakemake looks at the output patterns of each rule, sees whether they can create a file listed in `all`, and if so runs that rule with the corresponding wildcards. The `expand` statements are just wrappers around python `itertools` functions, usually `itertools.product`.


environment.yml
```diff
@@ -9,8 +9,13 @@ dependencies:
   - _libgcc_mutex=0.1
   - _openmp_mutex=4.5
   - _r-mutex=1.0.1
+  - aioeasywebdav=2.4.0
```
jjc2718, line 12: Are all of these packages that your code actually imports? I try to limit my `environment.yml` files to packages I'm directly importing in my Python code (or running from the command line like `snakemake` or `pytest`) to minimize clutter. My gut feeling is that this will improve portability, since fewer explicit requirements = more wiggle room to install packages/package versions that are specific to the user's system.

That's just my opinion though, I don't feel too strongly about this. It also means you have to remember to add things to this file by hand when you add new imports, but your tests/CI should break when this is out of sync so it shouldn't be a huge concern.

Snakefile
```diff
@@ -0,0 +1,69 @@
...
+               ),
+        # all_label_comparison_unsupervised
+        expand("results/all_labels.{unsupervised}.{supervised}.{dataset}.{seed}.tsv",
+               unsupervised=UNSUPERVISED,
+               supervised=SUPERVISED,
+               dataset=DATASETS,
+               seed=range(0,5)
+               ),
+
+rule pickle_compendium:
```
jjc2718, line 28: Is loading a pickle faster than loading from a .tsv file? Or is there another reason you're doing this?

Snakefile
```diff
@@ -0,0 +1,69 @@
+
+DATASETS, = glob_wildcards("dataset_configs/{dataset}.yml")
+SUPERVISED, = glob_wildcards("model_configs/supervised/{supervised}.yml")
+UNSUPERVISED, = glob_wildcards("model_configs/unsupervised/{unsupervised}.yml")
+
+wildcard_constraints:
+    # Random seeds should be numbers
+    seed="\d+"
```
jjc2718, line 8: Does this mean that whenever a variable called `seed` is defined in any rule, it has to be a number or the script will throw an error? Or does it only apply in certain places?

jjc2718 (approved):
Looks good! Just a few small comments. This is neat, I've used snakemake a bit in the past but not much - it's cool seeing what it can do.

The frequent PRs are totally fine! These don't take long to review so I don't mind at all. I should probably try to make mine smaller and more frequent TBH...

environment.yml
ben-heil: I prefer to include all the libraries (without builds) because it's automatic and more explicit. That being said, I imagine it's possible I'll run into an uninstallable environment file some day and change my opinion entirely

Snakefile
ben-heil: Right, when I tested it loading the tsv to a dataframe takes ~3 minutes and loading the pickled dataframe takes ~2 seconds

Snakefile
ben-heil: That's correct, the wildcarding likes to grab multiple fields at a time. Ensuring the seed is always a number helps reduce the number of rule priority commands I have to write

PR #15, opened 2020-09-09
Sepsis Benchmark

ben-heil:
This PR adds the logic to evaluate different models' performance on a single disease. It also runs and analyzes the benchmark on sepsis

Changes:
- Updated Snakefile to run the new benchmark script on sepsis 
- Threw together a notebook to visualize the results of the benchmark so far. The notebook will expand and change in the future, so I'm not super worried about its current form
- Caught a bug in `models.py` where batches of size 1 were squashed to scalars instead of single element 1-d vectors
- Made the scrip `single_label_prediction` to evaluate models trained on a single label


notebook/analysis/benchmark_results.ipynb
```diff
@@ -0,0 +1,668 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "['../../results/single_label.sepsis.PCA.pytorch_supervised.refinebio_labeled_dataset.3.tsv', '../../results/single_label.sepsis.PCA.pytorch_supervised.refinebio_labeled_dataset.2.tsv', '../../results/single_label.sepsis.logistic_regression.refinebio_labeled_dataset.1.tsv', '../../results/single_label.sepsis.PCA.logistic_regression.refinebio_labeled_dataset.1.tsv', '../../results/single_label.sepsis.PCA.logistic_regression.refinebio_labeled_dataset.4.tsv']\n"
```
jjc2718, line 39: Do all of the PCA preprocessed datasets here use the top 100 components? Do you have any plans to try other numbers of components?

notebook/analysis/benchmark_results.ipynb
```diff
@@ -0,0 +1,668 @@
...
+     },
+     "execution_count": 3,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "sepsis_metrics = None\n",
+    "for path in in_files:\n",
+    "    if sepsis_metrics is None:\n",
```
jjc2718: I think you could probably get rid of some of this duplicate code in this `if` statement, if you want to. Maybe you could initialize `sepsis_metrics` as an empty dataframe outside the loop (instead of `None`)? Then I think you can get rid of the `if` and just use the `else` (I'm pretty sure concatenating to an empty dataframe should just output the second argument, although you should probably test this).

Let me know if this is unclear.

notebook/analysis/benchmark_results.ipynb
```diff
@@ -0,0 +1,668 @@
...
+     "text": [
+      "<ggplot: (8771017871686)>\n"
+     ]
+    }
+   ],
+   "source": [
+    "plot = ggplot(sepsis_metrics, aes(x='supervised', y='accuracy', fill='unsupervised')) \n",
+    "plot += geom_violin()\n",
+    "plot += ggtitle('PCA vs untransformed data for classifying sepsis')\n",
+    "print(plot)"
```
jjc2718, line 259: Would you interpret this as the PCA and untransformed results being essentially the same? Or would you say the untransformed results are slightly better in the logistic regression case?

You might consider adding some kind of statistical test to answer this question (a t-test or nonparametric equivalent would probably be fine as a first pass, although [this post has some potential alternatives](https://stats.stackexchange.com/q/26271)).

saged/single_label_prediction.py
```diff
@@ -0,0 +1,150 @@
...
+        input_size = len(train_data.get_features())
+        output_size = len(train_data.get_classes())
+        print('output size: {}'.format(output_size))
+
+        if args.unsupervised_config is not None:
+            with open(args.unsupervised_config) as unsupervised_file:
+                unsupervised_config = yaml.safe_load(unsupervised_file)
+
+            # Initialize the unsupervised model
+            UnsupervisedClass = getattr(models, unsupervised_config['name'])
```
jjc2718: Do you want to delete the `name` attribute here too (since it's not being used in the class constructor, as far as I can tell)?

jjc2718 (approved):
Nice work! Just a few small comments/suggestions.

notebook/analysis/benchmark_results.ipynb
jjc2718: Actually I think code will be clearer than words in this case:

```python
sepsis_metrics = pd.DataFrame()
for path in in_files:
    new_df = pd.read_csv(path, sep='\t')
    model_info = path.strip('.tsv').split('sepsis.')[-1]
    model_info = model_info.split('.')
        
    if len(model_info) == 4:
        unsupervised_model = model_info[0]
        supervised_model = model_info[1]
    else:
        unsupervised_model = 'untransformed'
        supervised_model = model_info[0]
             
    new_df['unsupervised'] = unsupervised_model
    new_df['supervised'] = supervised_model
        
    sepsis_metrics = pd.concat([sepsis_metrics, new_df])
```
(I haven't tested this, so you definitely should if you decide to use it)

notebook/analysis/benchmark_results.ipynb
ben-heil: Yep, all the models use the same number of PCA components. I want to try other numbers of components/model architectures, etc, but I haven't come up with a good plan for hyperparameter optimization yet.

notebook/analysis/benchmark_results.ipynb
ben-heil: Good idea! Changed over to your version

notebook/analysis/benchmark_results.ipynb
ben-heil: Yeah, they look roughly the same to me. Good suggestion on the statistical tests, I'll definitely have to look more into it once I have more pieces of the benchmark up and running.

saged/single_label_prediction.py
ben-heil: Good call, the program now gets all the `name` variables from config with `pop` for consistency

PR #17, opened 2020-09-11
Fix housekeeping

ben-heil:
This PR follows up on the previous PR #16 and fixes the code coverage calculation

Changes:
- Code coverage calculation
- Added pytest-cov to environment file
- Removed an unnecessary class from `models.py`


jjc2718 (approved):
Looks good to me!

ben-heil:
Dear future Ben,

This commits reverted above were meant to be a part of a different PR. They also happen to not work, but that's a separate issue. Both problems were fixed by the revert commits above, allowing this PR to be closed.

PR #19, opened 2020-09-22
Test varying amounts of healthy data

### LINKED ISSUE #18: Nondeterminism Issue
The CV splits used in the `all_labels` analysis (and probably others) aren't the same between runs. I feel like this could be another set being cast to a list issue, but I'm unsure.

Code to check:
- get_cv_splits
- dataset constructors

ben-heil:
This PR measures the effects of reducing the amount of healthy training data on the models' accuracy. In both tb and sepsis increasing the amount of healthy data decreases the worst case performance, but doesn't seem to shift the mean very much.

Changes:
- Created `subset_label_prediction.py` to run the experiment
- Updated the snakefile to call `subset_label_prediction` 
- Fixed a determinism bug in `PytorchSupervised` models' initialization (closes #18)
- Moved some code in `single_label_prediction` to functions to be reused in `subset_label_prediction`


notebook/converted/benchmark_results.py
```diff
@@ -0,0 +1,271 @@
...
+# In[12]:
+
+
+print(sepsis_metrics[sepsis_metrics['healthy_used'] == 1])
+
+
+# In[13]:
+
+
+plot = ggplot(sepsis_metrics, aes(x='factor(healthy_used)', y='accuracy', ))
```
jjc2718, line 181: So here, you're taking subsets of the healthy training data, and keeping the sepsis training data the same? Are the test sets the same, or are you subsetting healthy data there too?

notebook/converted/benchmark_results.py
```diff
@@ -0,0 +1,271 @@
...
+print(plot)
+
+
+# In[20]:
+
+
+plot = ggplot(tuberculosis_metrics, aes(x='factor(healthy_used)', y='accuracy', fill='unsupervised')) 
+plot += geom_violin()
+plot += ggtitle('Effect of subsetting healthy data on prediction accuracy')
+print(plot)
```
jjc2718, line 261: To me, the "untransformed" distributions in this plot look pretty different than the distributions in the other two TB plots (these ones seem to get better as more healthy data is added, but in the other two plots that doesn't really seem to be the case, at least not as noticeably).

Shouldn't "untransformed" here be roughly the same as the first plot? Or is the setup here different somehow?

saged/models.py
```diff
@@ -19,6 +19,52 @@
...
+              embedded
+    Returns
+    -------
+    train_data: The embedded training data
+    val_data: The embedded validation data
+    unsupervised_model: The fitted version of the model
+    """
+    # Initialize the unsupervised model
+    unsupervised_model_type = unsupervised_config.pop('name')
+    UnsupervisedClass = globals()[unsupervised_model_type]
```
jjc2718: I was initially confused when I saw this, then I remembered that you have the comment later in this file explaining what you're doing here.

Could you write a short syntactic sugar function to do this? Call it `get_class_from_current_file()` or something like that, and just have it return this one-liner, then call it everywhere you need to get a class dynamically? The alternative would be to copy the comment everywhere you use this, but that seems like it might get a bit repetitive.

jjc2718 (approved):

notebook/converted/benchmark_results.py
ben-heil: The test sets are subsetted the same amount as the training sets, unless doing so would leave you with 0 samples. I did so to prevent the performance from being confounded by the label distribution between the training and test sets

saged/models.py
ben-heil: That's a good idea. I thought about copying the comment but chose not to since it was three lines long. Wrapping them in a function makes more sense so I went ahead and did so

notebook/converted/benchmark_results.py
ben-heil: Good eye! The distributions in the first plot are the sum of the untransformed and transformed distributions which washes out the signal. The distributions in the third plot are a combination of untransformed and transformed data split along a different axis (the supervised model used)

PR #20, opened 2020-10-13
Added pseudolabeling and a deep neural network model

ben-heil:
This PR adds a pair of new models including the first semi-supervised model!

The results imply that all the models perform roughly the same, or at least the data variance is overcoming the model variance. I'll look into that in future PRs

Changes:
- Added the logic to run semi-supervised methods to the snakefile and the benchmark scripts
- Added the DeepNetwork and Pseudolabel models to models.py
- Updated the results notebook with the results from running the benchmark on DeepNetwork and Pseudolabel


saged/models.py
```diff
@@ -811,3 +862,180 @@ def save_model(self, out_path: str) -> None:
...
+            label_counts = tune_dataset.map_labels_to_counts().values()
+
+            tune_baseline = max(label_counts) / sum(label_counts)
+            neptune.log_metric('tune_baseline', tune_baseline)
+
+        best_tune_loss = None
+
+        for epoch in tqdm(range(epochs)):
+
+            # Alpha (the ratio of pseudolabel loss to label loss to use
```
```suggestion
            # Alpha (the ratio of pseudolabel loss to label loss to use)
```

saged/models.py
```diff
@@ -811,3 +862,180 @@ def save_model(self, out_path: str) -> None:
...
+                progress = 0
+            alpha = progress * self.max_alpha
+
+            train_loss = 0
+            train_correct = 0
+            self.model.train()
+
+            # Create an iterator that returns a labeled batch and an unlabeled batch
+            if len(labeled_data) <= len(unlabeled_data):
+                # If there are more unlabeled samples than labeled samples, just stop iterating
```
jjc2718, line 968: Is this the default behavior of `zip` when one list is longer than the other? Might be worth noting that, although it makes intuitive sense.

saged/models.py
```diff
@@ -811,3 +862,180 @@ def save_model(self, out_path: str) -> None:
...
+                # once you've seen each labeled sample once
+                train_iterator = zip(train_loader, unlabeled_loader)
+            else:
+                # If the
```
jjc2718: Did you want to have a longer comment here?

saged/models.py
jjc2718, line 968: Also, I'm a bit confused by the comment here - isn't each of these a `DataLoader`? So wouldn't you be concerned about the number of batches in each, not the number of samples?

Probably not a big deal since you seem to be using the same batch size for the supervised and unsupervised loaders, but it might be good to make a note of this in case you vary batch sizes in the future.

jjc2718 (approved):
Code looks good to me! Exciting that you're starting to implement the semi-supervised models.

I have a few questions/comments on the results notebook (can't comment directly on it because the diff is too large):

* It seems like pseudolabeling gets rid of the long poorly performing tail for some of the subset experiments (e.g. for the TB analysis, somewhat for sepsis but not as noticeably). Is this because you're combining untransformed and PCA transformed results for the other methods (and when you look at the untransformed results later on the tails go away)? Or can you think of another explanation for this?
* ~~Is it correct to say that the results for all 4 of the supervised methods are approximately the same for these problems?~~ (just noticed you answered this in the PR description) Does this surprise you at all?

ben-heil:
> Code looks good to me! Exciting that you're starting to implement the semi-supervised models.
> 
> I have a few questions/comments on the results notebook (can't comment directly on it because the diff is too large):
> 
> * It seems like pseudolabeling gets rid of the long poorly performing tail for some of the subset experiments (e.g. for the TB analysis, somewhat for sepsis but not as noticeably). Is this because you're combining untransformed and PCA transformed results for the other methods (and when you look at the untransformed results later on the tails go away)? Or can you think of another explanation for this?
> * ~Is it correct to say that the results for all 4 of the supervised methods are approximately the same for these problems?~ (just noticed you answered this in the PR description) Does this surprise you at all?

Right, I didn't run PCA with pseudolabeling, so the pseudolabeling results in the figure you're referring to aren't truly comparable to the results of the other methods that were evaluated both with and without unsupervised learning

Yup, my take is that either 
1. The models all perform more or less identically despite having very different training/architectures, 
2. The variation across groups in the data (studies, platforms, blood types, or other bio/technical variation) is large enough that it drowns the signal, or
3. There's a programming error or some other factor I don't know about skewing the results

My working hypothesis is that 2 is the case, so I'm going to be digging into the inter-study/inter-platform performance in the next PR

saged/models.py
ben-heil: Yeah, that's not quite as descriptive of a comment as I'd like hahaha

Tbh I'm glad that comment got truncated. I had some clever plan for reducing the memory footprint of `itertools.cycle` by filling the shorter list with None, detecting None in the loop, and resetting the unlabeled dataloader. Unfortunately, it never got implemented (and also wouldn't have worked). Will update the comment and also fix the code

saged/models.py
ben-heil: Right, by zip stops after the shorter list. Will add a comment to note that

ben-heil: That's good to check, but `labeled_data` and `unlabeled_data` are the datasets with `train_loader` and `unlabeled_loader` are their corresponding loaders. They are created on lines 914 and 915 from the dataset passed in

PR #21, opened 2020-11-04
Batch effect correction

ben-heil:
This PR adds batch effect correction to the benchmark analyses. My goal for the PR was to see whether correcting for batch effects would reduce the variance in models' performance across seeds and data splits. The answer seems to be mostly no, though the models themselves have more clear performance/variance differences in corrected space.

There is also an interesting phenomenon that shows up where the 10% of healthy data subset performs better than the other subsetted splits. Maybe this is due to the decision to not subset if it leads to an empty validation set?

Changes:
- Added batch effect correction logic and argument to `subset_labels.py`
- Added batch effect correction runs to the Snakefile
- Updated the visualization notebook
- Added a function to wrap run batch effect correction on Dataset objects in `datasets.py`


notebook/analysis/benchmark_results.ipynb
```diff
@@ -1497,6 +1501,604 @@
...
...115DdnY2unbtihkzZqBBgwawWCyYMWMGsrOzQzLWwTp06IA1a9bg448/Rm5uLlq1ahVx1vrWW2/F7NmzwfM8iouLI7rtkCFDMH/+fIwfPx6DBg3CunXrsGrVqoDrTJkyBUeOHEGfPn1QWFiIH3/8ER988AHuvfdeAEC7du1w5513YuTIkXjggQfQvXt3eDwebN26FZ999hneeecdAMDEiRPRtWtXDBo0CHfccQe2bduGZ599NmzWX2ufqc3lIIQQkqaSPcuXEKIK7qLj9XrZfffdxxo0aMAyMzPZkCFD2FdffRXSMYYxxhYsWMDOOussJooiy8/PZ1deeSXbuXOn/vepU6eyZs2aMZ7nWcuWLfXLf/jhB3bppZcyh8PBMjIy2MCBA9nWrVsDtg2APfPMMyHjdblcrKCggL3yyisBlxvpIKPdftq0aaxt27ZMFEXWoEEDdtFFF7HXX39dv87vv//OLrroIuZwOFjz5s3ZM888w6ZMmcKcTqd+nXBddH766SfWu3dvlpmZyQCwhQsXMsbCd9eprXtRu3bt2BVXXBFyuRFPP/00a968OXM6nWzo0KHsww8/DHgM3n33XdavXz/WoEEDJkkSa9OmDZsyZQrzeDz6NhRFYS+88ALr2LEjE0WR5ebmsh49erBZs2YF3NeqVatYu3btmCRJrEuXLuyrr75iTqczpIvO7bffznr37h3V/hBCCDEPjjHGknd4QQgxs/vuuw/fffddQF/+dPHnn3+ibdu2WLlyJa699tpkDydmXq8XTZs2xTPPPFPjGgKEEELSAwX4hJCo7d+/H23atMGXX36Jc889N9nDiYsjR45g69atmDZtGn777Tf8/vvvNXakMZPFixdj+vTp+OWXX9JifwghhNSM2mQSQqLWqFEjLFq0CIcOHUr2UOLm3XffxQUXXIDt27fjjTfeCAmGZVkO6Esf/F+q4nkeCxYsoOCeEEJOAZTBJ4SQCPTp0yekXam/7du3o6ioKHEDIoQQQoJQgE8IIRH47bffUFpaWuPfO3XqFLaDDSGEEJIoFOATQgghhBCSRqgGnxBCCCGEkDRCAT4hhBBCCCFphAJ8QgghhBBC0ggF+IQQQgghhKQRCvAJIYQQQghJIxTgE0IIIYQQkkYowCeEEEIIISSNUIBPCCGEEEJIGvl/Af5e7HHJkeAAAAAASUVORK5CYII=\n",
```
jjc2718, line 1800: So are the means (or medians, whichever you want) of these distributions higher than when you don't use batch effect correction? Or are they similar, just with less variance? It might be nice to plot with/without correction side by side somewhere (for the same models), so it's easier to see the differences.

notebook/analysis/benchmark_results.ipynb
```diff
@@ -1497,6 +1501,604 @@
...
...Ea/M17RpU8ycORP//fefyUeXXn/9dXTu3Bnt27fHhAkTULNmTaSkpOCff/7B1atXsWzZMnh6eqJ+/fr47LPP4OvrC3t7e3z22Wfw9PTUO6LxsipVqmD37t3YsWMHvLy8ULZsWZOPagwdOhTR0dFQKpWIiIgw6bU9e/bE8uXLMWrUKISFheHgwYPYuHGjzjxRUVFITExEq1at4Ofnh3PnzuGPP/7AuHHjAACVKlXCiBEj8M4772D8+PFo2LAhsrKycPnyZezZswebN28GAEyaNAn169dHWFgY3nvvPVy/fh1ffPGFwaNCOcPVc/qCERUpUveaJqLnXh6llZ2dLT744APh6+sr3N3dRc+ePcWRI0f0RiQJIcSKFStEjRo1hKOjoyhevLjo1KmTuHnzpvb56dOni8DAQKFUKkVQUJB2+tmzZ8Ubb7whXFxchJubm+jcubO4fPmyzrIBiHnz5unlzcjIED4+PmLJkiU6040ZoZTz+hkzZoiKFSsKR0dH4evrK1q3bi1WrVqlnefKlSuidevWwsXFRZQuXVrMmzdPREVFCVdXV+08hkZpnT9/XjRv3ly4u7sLACI2NlYIYXj0Vl6j4ypVqiQ6dOigN90Yc+fOFaVLlxaurq4iPDxc/PnnnzrbYMuWLaJNmzbC19dXqFQqUb58eREVFSWysrK0y9BoNOLrr78W1atXF46OjsLLy0s0atRIfPXVVzrr2rhxo6hUqZJQqVSiXr164siRI8LV1VVvlNbw4cNF8+bNzWoPkdwphBBCunKLiOTsgw8+wN9//43du3dLHaXAXbt2DRUrVsSGDRvQo0cPqePkW3Z2NkqVKoV58+ahf//+UschsjgWPERktnv37qF8+fI4cOAA6tSpI3WcApGYmIjLly9jxowZuHTpEq5cuZLriCc5+e677zB79mxcuHDBJtpDZCoOSycis5UsWRIrV67U3prBFmzZsgVNmzZFXFwcvv/+e73iQK1W61wX5+X/rJVSqcSKFStY7FCRxSM8REQmaNWqld7lAV4UFxenc78yIrIOLHiIiExw6dIlJCcn5/p8zZo1DY6QIiJpseAhIiIim8c+PERERGTzWPAQERGRzWPBQ0RERDaPBQ8RERHZPBY8REREZPNY8BAREZHNY8FDRERENo8FDxEREdm8/we6bQm6bCSwLwAAAABJRU5ErkJggg==\n",
```
jjc2718, line 2043: Any particular reason you decided to look at a box plot here (rather than a violin plot)?

notebook/analysis/benchmark_results.ipynb
```diff
@@ -1497,6 +1501,604 @@
...
...ZsAhaH9kT3QuBR7cz0Uul0MhV6hJ8AFAapMwMTFRtT2VPbFByoO0X5Ik6z0AABqqX4NCoeBqseraxmIit6IoFduS5ZiLRCePRCIe6BioIFgzQIARwXc7BnM2gSa9DVJoUkA2k63YzoIFC9DTMwupzHrE5DMAAKnMM2ht7cCyZctCz8xxcLzesdcJfjweryDz6XS6gvSX47777quI3nd0dKCjwyCf3d3duPzyy3HllVcin887pi1vvPFGXH/99dbnz3zmM/jiF7/oo80J1wh+Z2cnmpubXdcBitFOtz6RGi8Is+1uaG1tdY3gd3Z2oq2tzXPbTTQ1NTleqAyouX/AIOnpiTTa2p06daFdxNj4WM31M5mMsS8Xgt/U1FTzvFVDMpk0CH7xPdDU1FT3GEZGRkB7XV5+vRR7hvfUXN+VnBHiab9uSCQSjgi+QAhkWa67rd27dyM2z0mQ5bkxbL1va911k8kkyqkNISRQ+wGD0Nm3xhiru60NL2+A4BJBlmbJkJpkbN++HUuWLHFdt5rLDCU08DFQSh0EPx6P19zWli1bEN8v6UpAhFYRiZ4kduzYgSOOOKLi+4mJYgJuRUdgbKu1tdX3cbS1tVnRbpNjtrW1uW5ny5YtUPIKYvvV7l/leXFQSUB/f79rhedYTHZKBot2jV1dXXUTvd1gLzll/lvtXJjBmXoSEbFVRD6fr3o+qSBAt0+0hHiW29raHBF8jTHEYu73kSAIJW/cOjBd15LJZMW2kskk9IJTuqOzPNrb3a99PcRisQoXpubm5optWVW03VykXEBiFGpBRWtrawXJP/HE4/D3+14AYBD8TO4FHHvsKnR2drpsiYODww/2OsGfM8dItNu5c6eVVLN161bXl4qJLVu2YMeOHTjllFNqbtssDlU+1XvFFVfgnHPOsT7HYrG6/tt2qJoGoWw6FjBe3rVs7qpF8CEYg5pabchms2B6peZ1cnIy0LT+5OSkM4Kvs7rnYHh42Ig6l0XOhDYRo8NDNdc3iY2z0JWxnaGhoUD2gJOTk6CkJHYYGxurewybt20G7akkBlKvhC1Pbqm5fmm635mkmkqlfN0/JsbHx2GXQVMwT9t65vlnK6pgygsS6Ht1K/r7+2sSrEwmU0bMDFIepP2AEc0r106PjIxUjb6pqornn3se3adURpAJJYgvS+CRRx5xJM7b4V6xl0FneuBj
```
jjc2718, line 2080: This plot is essentially the same in cell 24, but with batch correction (and without the semi-supervised method), right?

If so, it's interesting how different this plot is. Particularly for logistic regression - by eye it looks to me like performance has gotten worse for logistic regression, but not so much for the deep learning methods (although they still have higher variance).

jjc2718, line 2080: Also, small question - is "pytorch_supervised" here the three-layer neural net from `saged/models.py`? It looks like it's safe to assume that this is the case, but it's a bit ambiguous since your "deep_net" also seems to use the `PytorchSupervised` class downstream.

jjc2718 (commented):
Nice work! Code all looks good to me, just a few small questions about the analysis/interpretation.

jjc2718 (approved):
Sorry, meant to approve and clicked the wrong button.

As far as this question:

> There is also an interesting phenomenon that shows up where the 10% of healthy data subset performs better than the other subsetted splits. Maybe this is due to the decision to not subset if it leads to an empty validation set?

This makes intuitive sense to me - if the data being used in the 10% subset experiments is much different than for the other experiments, the results wouldn't be as surprising. Maybe you could check which samples (or studies, whichever makes sense) are being included and not included in the 10% experiments, and look into those? (That can be a separate issue or PR, feel free to merge this PR when you want)

notebook/analysis/benchmark_results.ipynb
ben-heil: I think the means are higher when I'm not using batch correction. I don't have a strong theory about why that is. Maybe there is some degree of biological signal confounded with study structure that gets removed by limma

I agree they should be plotted side by side, will do in a future PR

notebook/analysis/benchmark_results.ipynb
ben-heil: Not really, I was just curious where the means were so I converted this one to a boxplot

notebook/analysis/benchmark_results.ipynb
ben-heil: Exactly

Yeah, I was surprised at how different it looks too

ben-heil: Right, pytorch supervised is the shallow network. It's named that way because it was the first model using PytorchSupervised, but I need to clarify that at some point

ben-heil:
> Sorry, meant to approve and clicked the wrong button.
> 
> As far as this question:
> 
> > There is also an interesting phenomenon that shows up where the 10% of healthy data subset performs better than the other subsetted splits. Maybe this is due to the decision to not subset if it leads to an empty validation set?
> 
> This makes intuitive sense to me - if the data being used in the 10% subset experiments is much different than for the other experiments, the results wouldn't be as surprising. Maybe you could check which samples (or studies, whichever makes sense) are being included and not included in the 10% experiments, and look into those? (That can be a separate issue or PR, feel free to merge this PR when you want)

Yep, there's a number of interesting things going on that are the subject of my next PR!

PR #22, opened 2020-11-18
More Subsetting Methods

ben-heil:
This PR contains methods used to correct for the imbalanced label distributions causing the v-shaped curves seen in https://github.com/greenelab/saged/pull/21. 

The visualizations for the subsetting methods can be found in `subsetting_visualizations.ipynb`. If you get the chance, please double check my math in `determine_subset_fraction`. I'm pretty sure it's correct now, and the test cases no longer break, but it took me enough tries that I'm wary of it still being subtly wrong


Snakefile
```diff
@@ -107,6 +107,25 @@ rule all:
                dataset=DATASETS,
                seed=range(0,NUM_SEEDS)
                ),
+        # subset_all sepsis batch effect corrected
+        expand("results/subset_all.sepsis.{supervised}.{dataset}.{seed}.be_corrected.tsv",
```
jjc2718, line 111: Do you want to run this experiment for TB as well?

saged/utils.py
```diff
@@ -365,3 +365,40 @@ def deterministic_shuffle_set(set_: set) -> List[Any]:
...
+
+    Returns
+    -------
+    subset_fraction: The fraction of positive or negative (determined by the calling code) samples
+                     to remove
+    """
+    train_disease_fraction = train_positive / (train_negative + train_positive)
+    val_disease_fraction = val_positive / (val_positive + val_negative)
+
+    # If train ratio is too high, remove positive samples
```
jjc2718, line 393: This math looks correct to me! (And the logic makes sense as well, at least as far as I can tell)

saged/keep_ratios.py
```diff
@@ -0,0 +1,226 @@
...
+    train_disease_fraction = train_positive / (train_positive + train_negative)
+    val_disease_fraction = val_positive / (val_positive + val_negative)
+
+    subset_fraction = utils.determine_subset_fraction(train_positive,
+                                                      train_negative,
+                                                      val_positive,
+                                                      val_negative)
+
+    # If train ratio is too high, remove positive samples
+    if train_disease_fraction > val_disease_fraction:
```
jjc2718, line 46: Could this go in the `determine_subset_fraction` function, or did you already think about that and decide to put it here?

I guess I'm of two minds about it being here: if the logic for manipulating the datasets was in the `determine_subset_fraction` function you wouldn't need these inequalities in multiple places (would be nice to have a single point of truth for them), but I also get wanting to not have your utility function do too much (if you move this logic in there you'll have to pass the dataset objects as arguments to the function, which might make things messier, and then you'd also probably need to change the name of the function since it wouldn't really describe everything it's doing, etc).

Up to you -- I just tend to be wary of duplicating bits of logic like this across multiple files, since then you have to remember where they all are if you want to change them in the future.

jjc2718 (approved):
Looks good! This is a neat analysis -- glad to see that your latest results are closer to expectations.

A few questions/comments on `subset_visualizations.ipynb` (looks like the diff is too large to make inline comments):

* Intro text: by "confounded results" you mean performance is artificially inflated in the small subset percentages, correct? Might be clearer to just say that (confounded results can mean a lot of different things in different contexts) although linking to the plot helps.

* In the grouped/faceted point plot (cell 21), it might be good to label the groups. Looks like they're different random seeds?

* Subsampled TB data experiments (cells 29-37): so subsampling to balance the label distributions seems to get rid of the V shape, but were you expecting to see more of a clear performance improvement as the amount of training data increases? Looks like in sepsis adding data increases performance somewhat, but not so much in TB - does that match your impression/expectation?

* Trend line plots for each model type (cells 54 and 62): interesting (although not surprising) that logistic regression seems to be the least sensitive to added training data. Why do you think `pytorch_supervised` (which if I remember correctly is a 3-layer MLP) seems to outperform `deep_net` (more than 3 layers IIRC) fairly consistently? Do you think more data would help (the trend lines you plotted seem to suggest it wouldn't, but the high counts are a bit undersampled so maybe that's misleading)?

ben-heil:
Ultimately I don't intend to mess with the `visualize_subsetting` notebook much since it's more of a debugging notebook then actual results. The end analysis `keep_ratios` will be added to the main `benchmark_results` notebook in my next PR. I mostly just added the notebook to the main branch as a description of one failure state of study-stratified data splitting

> * Intro text: by "confounded results" you mean performance is artificially inflated in the small subset percentages, correct? Might be clearer to just say that (confounded results can mean a lot of different things in different contexts) although linking to the plot helps.

It's not strictly that the performance is inflated in the small subset percentages, that's just how it manifests here. The two signals (performance bonus of data and performance malus of the label difference gap) are overlayed on top of each other, making it hard to determine what is what.

> * In the grouped/faceted point plot (cell 21), it might be good to label the groups. Looks like they're different random seeds?
> * Subsampled TB data experiments (cells 29-37): so subsampling to balance the label distributions seems to get rid of the V shape, but were you expecting to see more of a clear performance improvement as the amount of training data increases? Looks like in sepsis adding data increases performance somewhat, but not so much in TB - does that match your impression/expectation?

Yeah, this was really surprising to me! Based on the accuracy vs label_gap vs disease_count figure at the bottom it seems like the tb models are just guessing negative class each time. Definitely something to look into further and hopefully fix in the future.

> * Trend line plots for each model type (cells 54 and 62): interesting (although not surprising) that logistic regression seems to be the least sensitive to added training data. Why do you think `pytorch_supervised` (which if I remember correctly is a 3-layer MLP) seems to outperform `deep_net` (more than 3 layers IIRC) fairly consistently? Do you think more data would help (the trend lines you plotted seem to suggest it wouldn't, but the high counts are a bit undersampled so maybe that's misleading)?

My guess would be that there isn't enough data to train a medium-size network, so the increase in variance caused by the extra layers overshadows the decrease in bias. I'm curious of what the results would be if I tried a monster-sized network on gene expression data, but that's a project for the medium future.

I imagine with enough data the deeper net would perform the best, but we only have a few thousand samples in the analyses

Snakefile
ben-heil: It probably won't be necessary since the `keep_ratios` experiment should be a strictly better gauge of performance. I'll go ahead and add TB to the snakemake file though so it gets run at some point

saged/utils.py
ben-heil: Yay! Sixth time's the charm!

PR #25, opened 2020-11-21
Better metrics and better label encoding

### LINKED ISSUE #23: Label encoding can change if I'm not careful
Right now labels are encoded at dataset creation time using sklearn.LabelEncoder, which sorts the labels before encoding them [method one](https://github.com/scikit-learn/scikit-learn/blob/0fb307bf3/sklearn/preprocessing/_label.py#L37), [method 2](https://github.com/scikit-learn/scikit-learn/blob/0fb307bf3/sklearn/preprocessing/_label.py#L58). This means that for classification where the labels are the same between datasets, the labels' encodings should also be the same. However, if there are more labels in the training set than validation set, the validation might be comparing apples to oranges. 

This might be the cause of the bad performance in the `all_labels` analysis, and definitely is something to make sure to get right. I'll be adding the logic to account for this (`LabeledDataset.get_label_encoder()` and `LabeledDataset.set_label_encoder()`) to the next PR

ben-heil:
This PR implements the logic to fix https://github.com/greenelab/saged/issues/23 going forward, and adds balanced accuracy and f1_score to the `keep_ratios.py` script. Shout outs to past me for making the evaluation function compatible with sklearn metrics.

You don't need to spend much time on the notebook, the only big difference in behavior from implementing the new metrics is visible at the bottom, where the accuracy and balanced accuracy results are different for TB.

Changes:
- Updated benchmark results notebook with balanced acc plots for keep_ratios
- Improved label encoding in `datasets.py`
- Added metrics in `keep_ratios.py`


saged/keep_ratios.py
```diff
@@ -208,19 +227,20 @@ def subset_to_equal_ratio(train_data: datasets.LabeledDataset,
...
-                             supervised_train_sample_names,
-                             supervised_val_sample_names,
-                             supervised_train_sample_counts,
-                             subset_percents,
-                             ):
-            out_file.write(f'{accuracy}\t{train_study_str}\t{train_sample_str}\t')
-            out_file.write(f'{val_sample_str}\t{supervised_train_samples}\t{percent}\n')
+        # Write header
+        out_file.write('accuracy\tbalanced_accuracy\tf1_score\ttrain studies\ttrain samples\t')
+        out_file.write('val samples\ttrain sample count\tfraction of data used\n')
```
jjc2718: I think you should just be able to do this here:
```suggestion
        out_file.write('accuracy\tbalanced_accuracy\tf1_score\ttrain studies\ttrain samples\t'
                       'val samples\ttrain sample count\tfraction of data used\n')
```
(the strings should automatically be concatenated).

saged/keep_ratios.py
```diff
@@ -196,8 +208,15 @@ def subset_to_equal_ratio(train_data: datasets.LabeledDataset,
             supervised_model.free_memory()
 
             accuracy = sklearn.metrics.accuracy_score(predictions, true_labels)
```
jjc2718: Should this be:
```suggestion
            accuracy = sklearn.metrics.accuracy_score(true_labels, predictions)
```
(I don't think it matters to the output since accuracy should be symmetric, but probably good to be consistent with the other metrics)

notebook/analysis/benchmark_results.py
```diff
@@ -0,0 +1,591 @@
...
+tb_stat_df = create_dataset_stat_df(tb_metrics, 
+                                    sample_to_study, 
+                                    sample_metadata, 
+                                    sample_to_label,
+                                    'tb')
+
+tb_stat_df.tail(5)
+
+# %%
+ggplot(tb_stat_df, aes(x='train_val_diff',
```
jjc2718, line 583: Do you think this plot would change at all with balanced accuracy on the Y axis?

jjc2718 (approved):
Looks good to me!

notebook/analysis/benchmark_results.py
ben-heil: I updated the plot, and it looks about like I would expect. The overall trend changes a little bit, as seen in the plot afterwards, but mainly it just gets rid of the accuracy 1 runs where the model guesses the positive class everytime for LR. The deepnet still runs into that, but it's split between accuracy 1 and accuracy .5 depending on which class it decides to guess each time

PR #28, opened 2020-11-25
Small subset

ben-heil:
This PR finds the performance crossover point between a three layer neural network and logistic regression by looking at small training datasets.

Interestingly, we find that tb increases to a point then flatlines, while sepsis seems to take more data to saturate. Also it looks like the training of `deep_net` is very broken, which is unsurprising given past results.

Changes:
- Normal snakefile changes to add `small_subsets.py` 
- Update benchmark results notebook with figures showing the crossover point
- Update the model code to patch the case where the training set is too small to have a good tune set. There is probably a more elegant way to handle this, but I'll do that when I implement early stopping
- Implement `small_subsets.py`


saged/models.py
```diff
@@ -537,6 +538,13 @@ def fit(self, dataset: LabeledDataset) -> "PytorchSupervised":
         train_dataset, tune_dataset = dataset.train_test_split(train_fraction=train_fraction,
                                                                train_study_count=train_count,
                                                                seed=seed)
+        # For very small training sets the tune dataset will be empty
```
jjc2718, line 541: Do you have to hold out entire studies for your tune dataset? I get why that's important for your test data (you want to evaluate if your model generalizes to unseen studies), but for your tune data maybe it's okay to split up studies, at least in the case when your training set is small.

Just an idea, not necessary to change anything now.

saged/small_subsets.py
```diff
@@ -0,0 +1,254 @@
...
+healthy and diseased gene expression with equal label distributions in the train and val sets."""
+import argparse
+
+import sklearn.metrics
+import yaml
+
+from saged import utils, datasets, models
+
+
+def subset_to_equal_ratio(train_data: datasets.LabeledDataset,
```
jjc2718, line 12: This is similar to the function in `saged/keep_ratios.py`, right? Could you combine them into a single function?

notebook/analysis/benchmark_results.py
```diff
@@ -581,11 +588,107 @@ def return_unlabeled():
...
+print(plot)
+
+# %%
+plot = ggplot(tb_metrics, aes(x='factor(train_count)', y='balanced_accuracy', fill='supervised')) 
+plot += geom_boxplot()
+plot += ggtitle('TB Dataset Size vs Models (equal label counts)')
+print(plot)
+
+# %%
+plot = ggplot(tb_metrics, aes(x='train_count', y='balanced_accuracy', color='supervised'))
```
jjc2718, line 690: You could try plotting this with error bars (e.g. standard error) to emphasize that the crossover isn't really significant, if you want. (At least I assume it won't be, looking at the data points)

jjc2718 (approved):
Looks good to me!

saged/models.py
ben-heil: That's a good idea, but given the differences between studies I imagine I would end up overfitting if I did early stopping using the tune loss. I could probably do the splitting differently though. Maybe I should just designate a single study at the beginning of the run to be a tuning set *shrug*. 

I'll open an issue and add your comment to it for thinking about in the future

saged/small_subsets.py
ben-heil: Right, it's an updated version of the one in saged/keep_ratios.py. When I refactor I'm moving this one to `utils.py` and calling changing both benchmarks to just call the utils version

notebook/analysis/benchmark_results.py
ben-heil: I think `geom_smooth` is supposed to plot confidence intervals by default but didn't here for some reason. Maybe the fact that the x axis isn't really continuous?

PR #30, opened 2020-12-18
Non be corrected

ben-heil:
I forgot to put together a PR for the non batch effect corrected analyses so this is kind of an agglomeration of two different PRs. No rush though, I won't be able to check the PR until January anyway

The first part compares visualizes the difference between batch effect corrected and non-corrected data for training models.
Without BE correction, everything works super well on sepsis for reasons that I am unclear about. BE correction makes no difference on tb performance though. Maybe sepsis has a strong signal that's confounded with batch effects?

The second part runs the `keep_ratios` benchmarking script on lupus and visualizes the results. Lupus behaves similarly to TB in that be correction doesn't make much of a difference on performance and that they both have weird outliers around .9-1 balanced accuracy. I need to check for a data leak or duplicated samples for those.

The main things to look at are the difference between the batch effect corrected and uncorrected plots (cells 67-82) and the lupus plots (83 onward).


notebook/analysis/benchmark_results.py
```diff
@@ -688,7 +693,266 @@ def return_unlabeled():
...
+plot += ggtitle('Sepsis Datset Size by Model (equal label counts)')
+print(plot)
+
+# %%
+plot = ggplot(sepsis_metrics, aes(x='train_count', y='balanced_accuracy', color='supervised')) 
+plot += geom_smooth()
+plot += geom_point(alpha=.2)
+plot += geom_hline(yintercept=.5, linetype='dashed')
+plot += ggtitle('Sepsis Crossover Point')
+plot
```
jjc2718, line 751: This is interesting! It looks like performance overall is much better without batch correction, and there's no crossover point anymore (logistic regression seems to always do better).

Maybe this could be similar to what Alex's work showed about compendia where many studies were combined? Although you're seeing improved performance even for very small training sets (probably only a few studies), so that doesn't seem like it would explain the whole difference.

Are you just using `limma` for batch effect correction? I wonder if you'd see the same performance difference if you used another method, or if this is particular to `limma`.

notebook/analysis/benchmark_results.py
```diff
@@ -688,7 +693,266 @@ def return_unlabeled():
...
+lupus_metrics['supervised'] = lupus_metrics['supervised'].str.replace('pytorch_supervised', 'three_layer_net')
+lupus_metrics
+
+# %%
+plot = ggplot(lupus_metrics, aes(x='train_count', y='balanced_accuracy', color='supervised')) 
+plot += geom_smooth()
+plot += geom_point(alpha=.2)
+plot += geom_hline(yintercept=.5, linetype='dashed')
+plot += ggtitle('lupus Crossover Point')
+plot
```
jjc2718, line 919: Interesting that the accuracy = 1 points go away in lupus when you correct for batch effects, but they seem to still be there in TB with or without batch effect correction.

notebook/analysis/benchmark_results.py
```diff
@@ -688,7 +693,266 @@ def return_unlabeled():
...
```
jjc2718, line 958: A bit of an unrelated question, but are you doing any sort of normalization/standardization on the input gene expression data? It looks like refine.bio does quantile normalization by default - you could try standardizing (taking z-scores of) each gene instead, although I'm not sure how much difference this would make vs. quantile normalization.

This might be more trouble than it's worth, though, if the data took a long time to download.

jjc2718 (approved):
Looks good! These are definitely surprising results.

I wouldn't have expected batch effect correction to decrease performance (at least in the sepsis case) - my guess would be, like you said, that it's removing signal which is confounded with batch somehow. Interested to see where you go from here.

notebook/analysis/benchmark_results.py
ben-heil: > Maybe this could be similar to what Alex's work showed about compendia where many studies were combined? Although you're seeing improved performance even for very small training sets (probably only a few studies), so that doesn't seem like it would explain the whole difference.

It could be? Though like you said I'm probably not working with enough studies for that to be the main factor.

I'm currently just using limma. A super abridged unsupervised comparison had them working roughly the same (https://github.com/greenelab/saged/blob/master/notebook/analysis/batch_effect_correction.ipynb). It's something to check out in the future, though my intuition is that different correction methods would lead to different points along the [batch effects dominate <-> signal removed] scale

notebook/analysis/benchmark_results.py
ben-heil: The accuracy ~ 1 points tend to come from the same data split (each run with the same data split yields 10 points), so it's not super surprising that that happens. My guess is that there's some duplication in the data or data leakage somewhere.

notebook/analysis/benchmark_results.py
ben-heil: I'm currently just using the quantile normalization from refine.bio. It may be worth trying other methods like z-scoring or [class-specific quantile normalization](https://www.nature.com/articles/s41598-020-72664-6.pdf).

PR #32, opened 2021-01-19
Wrote EDA notebook to compare the similarity between samples/studies

ben-heil:
Previous experiments have had some strange results, with models occasionally performing abnormally well (or badly) on the out of sample set. This PR looks for duplicate samples, weird studies, etc.

Ultimately nothing stands out as causing problems (though I need to make sure the beadchip studies are staying removed).

### DIFF
--- a/notebook/debugging/data_similarity.ipynb
+++ b/notebook/debugging/data_similarity.ipynb
@@ -1,0 +1,1306 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Data Similarity\n",
+    "Previous experiments have had some strange results, with models occasionally performing abnormally well (or badly) on the out of sample set. To make sure that there are no duplicate samples or abnormally similar studies, I made this notebook"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import json\n",
+    "\n",
+    "import matplotlib.pyplot as plt\n",
+    "import numpy as np\n",
+    "import pandas as pd\n",
+    "import yaml\n",
+    "from plotnine import *\n",
+    "from sklearn.metrics.pairwise import euclidean_distances\n",
+    "\n",
+    "from saged import utils, datasets, models"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Load the data"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "dataset_config_file = '../../dataset_configs/refinebio_labeled_dataset.yml'\n",
+    "label = 'tb'\n",
+    "negative_class = 'healthy'\n",
+    "batch_correction_method = 'limma'"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "dataset_config_str = \"\"\"name: \"RefineBioMixedDataset\"\n",
+    "compendium_path: \"../../data/subset_compendium.pkl\"\n",
+    "metadata_path: \"../../data/aggregated_metadata.json\"\n",
+    "label_path: \"../../data/sample_classifications.pkl\"\n",
+    "\"\"\"\n",
+    "dataset_config = yaml.safe_load(dataset_config_str)\n",
+    "\n",
+    "dataset_name = dataset_config.pop('name')                                   \n",
+    "MixedDatasetClass = datasets.RefineBioMixedDataset                                \n",
+    "\n",
+    "all_data = MixedDatasetClass.from_config(**dataset_config)                  \n",
+    "labeled_data = all_data.get_labeled()                                       \n",
+    "labeled_data.subset_samples_to_labels([label, negative_class])              \n",
+    "unlabeled_data = all_data.get_unlabeled()                                   "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "metadata_path = '../../data/aggregated_metadata.json'\n",
+    "metadata = None\n",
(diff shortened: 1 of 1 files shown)


notebook/debugging/data_similarity.ipynb
```diff
@@ -0,0 +1,1306 @@
...
+   ],
+   "source": [
+    "plt.hist(nonzero_distances, bins=20)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Distribution looks bimodal, probably due to different platforms having different distances from each other?"
```
jjc2718, line 1561: Could you color/facet your plot by platform? Or is there too many of them for this to be informative?

notebook/debugging/data_similarity.ipynb
```diff
@@ -0,0 +1,1306 @@
...
+   "cell_type": "code",
+   "execution_count": 20,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# There is almost certainly a vectorized way of doing this but oh well\n",
+    "distances = []\n",
+    "first_samples = []\n",
+    "second_samples = []\n",
+    "for row_index in range(sample_distance_matrix.shape[0]):\n",
```
jjc2718, line 1662: You could probably use something like [np.nonzero](https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html) to get all the nonzero indices, then index into the sample arrays using those, if you want to.

Not a big deal, though, if what you have doesn't take too long to run.

notebook/debugging/data_similarity.ipynb
```diff
@@ -0,0 +1,1306 @@
...
+   ],
+   "source": [
+    "means_df.sort_values(by='difference')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "These results indicate that most of the data is behaving as expected (the distance between pairs of samples from different studies is less than the distance between pairs of samples within the same study).\n",
```
jjc2718, line 2481: You wouldn't necessarily expect the distance to be less between different studies, would you? Just somewhere close to 0 (so you're really looking for outliers in the negative direction here)?

notebook/debugging/data_similarity.ipynb
```diff
@@ -0,0 +1,1306 @@
...
+    "\n",
+    "The outliers are mostly bead-chip, which makes sense (though they shouldn't be in the dataset and I'll need to look more closely at that later). The one exception is SRP049820 which is run on an Illumina Genome Analyzer II. Maybe it's due to the old tech?"
```
jjc2718, line 2483: Are these outliers the same as the holdout data in the abnormally performing cases that you mentioned at the beginning of the notebook?

notebook/debugging/data_similarity.ipynb
```diff
@@ -0,0 +1,1306 @@
...
+       "<ggplot: (8791901875637)>"
+      ]
+     },
+     "execution_count": 37,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "plot = ggplot(means_df, aes(x='study_name', y='difference')) \n",
```
jjc2718, line 2330: I wonder if you could correct this for platform variance somehow - it seems like some platforms are going to tend to have lower or higher in-study mean pairwise distances just by virtue of their platform, not because of study-specific variation (unless you're trying to get at differences between platforms, but it seems like maybe you want to detect outlying studies without the potential confounding effect of platform).

Here are a few ideas, up to you whether to pursue them or not:

* Could you regress out the platform (i.e. something super simple like what's described in [the intro to this paper](https://www.biorxiv.org/content/10.1101/2020.08.17.255034v1.abstract)), then do the same distance analysis? Or are there many different platforms such that each study is essentially a separate platform?
* Could you correct for changing variance between studies somehow? Maybe you could use something like the average [Mahalanobis distance](https://en.wikipedia.org/wiki/Mahalanobis_distance) comparing each point in the study with the mean of the study vs. the mean of the whole dataset for this, although I haven't thought this through in a super precise way.

jjc2718 (approved):
Looks good to me! I made some conceptual comments/tried to give you a few ideas on possible future directions, but I think I'm mostly convinced there's nothing too weird going on.

notebook/debugging/data_similarity.ipynb
ben-heil: That's a good idea, but I can't think of a clean way to do platforms since the distances correspond to pairs of samples. Also there's 50 million entries and my computer started swapping when I tried to do something similar with the distances at the end

notebook/debugging/data_similarity.ipynb
ben-heil: That's a good idea, I tried it out and ran out of memory though hahaha

notebook/debugging/data_similarity.ipynb
ben-heil: I would expect the distance between samples within the same study to be greater than the distance of two samples from different studies based on [these results](https://github.com/greenelab/whistl/blob/master/notebook/data_exploration/compendium_eda.ipynb). Very positive outliers would also be interesting because they would indicate that a study's samples are very similar to each other

notebook/debugging/data_similarity.ipynb
ben-heil: I didn't think about that, but it may be the case. I need to look into it when I get VPN access back

notebook/debugging/data_similarity.ipynb
ben-heil: That's a good point, and may be something I was accidentally already doing. The initial results showed the effects after batch effect correction where I corrected by platform. I'm currently running the same analysis without be correction to see whether things change

jjc2718: Got it! Yeah, I would imagine that BE correction would fix different variances per platform (AFAIK most methods are essentially regressing out the batch/platform like in my first bullet point, often with some other bells and whistles) but I'm not entirely sure. Comparing against the same analysis with no correction will definitely help clarify what's going on.

ben-heil: The distributions look mostly the same when not batch effect correcting, except cross-study difference is much greater across the board

PR #33, opened 2021-03-29
Imputation

ben-heil:
This PR looks at whether increasing the amount of data available to an imputation model increases its ability to predict gene expression in other studies.

The major changes are in the `models.py` file's imputation training logic. Besides that, the new notebook shows that increasing data decreases MSE in the validation set.


saged/impute_expression.py
```diff
@@ -0,0 +1,131 @@
+"""
+This benchmark compares the performance of different models in learning to differentiate between
```
jjc2718: Is this comment accurate, or did you copy it from another script? I didn't think you were classifying healthy vs. diseased in these experiments, unless I'm misunderstanding.

saged/impute_expression.py
```diff
@@ -0,0 +1,131 @@
...
+                        help='The number of splits to use in cross-validation',
+                        type=int,
+                        default=5)
+    parser.add_argument('--batch_correction_method',
+                        help='The method to use to correct for batch effects',
+                        default=None)
+
+    args = parser.parse_args()
+
+    # TODO
```
jjc2718, line 37: You could move these TODOs to issues (although I think this is a bit up to personal preference - some people like to use issues mostly for bugs, and keep future ideas/extensions elsewhere, up to you).

saged/models.py
```diff
@@ -368,6 +396,331 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        FileNotFoundError if out_path isn't openable
+        """
+        torch.save({
+                    'model_state_dict': self.model.state_dict(),
+                    'optimizer_state_dict': self.optimizer.state_dict(),
+                    },
+                   out_path
+                   )
+
+    def mask_input_(self,
```
jjc2718, line 534: Just so I understand what you're doing - this function masks varying proportions of the input, so you can evaluate the effect of having more or less data in your train set?

If so, don't you want to remove these data points entirely from your training set, rather than setting them to 0? Or does setting the inputs to 0 effectively remove them? (I'm super out of practice with Pytorch so I could be missing something basic)

saged/utils.py
```diff
@@ -50,6 +50,24 @@
...
+    fraction_zeros: The fraction of elements in the resulting tensor that should
+                    be zero (the rest will be set to ones)
+
+    Returns
+    -------
+    mask: The mask tensor of shape `shape`
+    '''
+    # Get [0,1] random numbers, and set all greater than fraction_zeros to 1
+    # (and set all others to 0)
+    return torch.rand(shape) > fraction_zeros
```
jjc2718: This won't necessarily give you exactly `fraction_zeros` zeros though, right? In effect, I think you're making a bunch of draws from a Bernoulli distribution with parameter `p = fraction_zeros`, which in the limit of many draws should give you `fraction_zeros` zeros, but on smaller samples there could be a considerable amount of variation.

Probably doesn't matter much in practice, especially if you're working with sample sizes in the hundreds/thousands or larger, but good to keep in mind and maybe put in the function documentation.

notebook/converted/imputation_viz.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+# In[6]:
+
+
+metrics = pd.DataFrame()
+for path in in_files:
+    new_df = pd.read_csv(path, sep='\t')
+    
+    #print(new_df.head())
```
jjc2718: Looks like this was for debugging (so maybe you can remove it)?

notebook/converted/imputation_viz.py
```diff
@@ -0,0 +1,128 @@
...
+    
+metrics = metrics.reset_index()
+metrics['trial'] = metrics.index // 10
+metrics
+
+
+# In[7]:
+
+
+ggplot(metrics, aes(x='train sample count', y='val_loss', color='factor(trial)')) + geom_line()
```
jjc2718, line 84: What is the validation loss here? Is this MSE, or MAE or something like that?

I usually like to calculate both MSE and RMSE when I do regression/prediction of a continuous output. I generally lean toward plotting RMSE in most cases since it's in the same units as the input (slightly more interpretable at a glance), but that's just me.

jjc2718 (approved):
Interesting stuff! Just a few questions/comments.

More generally, just to check my own understanding: in the context of this PR/project, "imputation" = the following:

* Hide a subset of gene expression features in the training set
* Train a model using the hidden expression values as labels, and the non-hidden values as training data
* Evaluate the model on the same hidden/non-hidden expression values for some held-out samples (or studies)

Is that the general idea, or am I missing anything?
 
Also, are the included studies from all the disease types you were looking at previously, or only some subset of those?

saged/impute_expression.py
ben-heil: Good catch, I would have sworn that I updated it, but it must have been a different file

saged/impute_expression.py
ben-heil: Will do eventually, `impute_expression` is still a work in progress so I'm fine with leaving it here for now

saged/models.py
ben-heil: The masking function does genewise dropout, not samplewise dropout. The idea is that an easy self-supervised learning task is to impute or denoise data, like in masking tasks in transformers. There is probably also a ton of related literature that I haven't read yet from single cell expression because their data does this to itself

Training on masking 
1) gives us an easy metric to evaluate differences in models/data with and
2) Can hopefully be used as a pretraining step in the future

saged/utils.py
ben-heil: That's correct, I'll update the documentation to be more precise

notebook/converted/imputation_viz.py
ben-heil: It's the sum of the MSE for all batches. I should definitely convert it to something more interpretable eventually, but I'm not entirely sure what that would be (per gene error? per sample?)

PR #34, opened 2021-04-16
Pretraining via Imputation

### LINKED ISSUE #29: Feature: Early Stopping
PytorchSupervised models should implement early stopping at some point. The current train/tune split doesn't work great though, maybe use ideas from https://github.com/greenelab/saged/pull/28#discussion_r532795745 ?

ben-heil:
To determine whether the noise that I've been seeing in model training so far might be alleviated by starting with good data-based initializations, I built a pretraining pipeline. The pipeline masks random genes by setting their expression values to zero, then trains a model to predict the true gene expression. This masking task is inspired in part by denoising autoencoders, and in part by the masking pretext tasks used to train transformers.

Ultimately, while the learning curves on the masking task look fine, when these models are transferred to disease classification you still see huge error bars in the sepsis case and disconnected train/val set accuracy in the tb case.

The new infrastructure for the pipeline can be found in datasets.py, models.py, Snakefile, and imputation_pretraining.py. The results are visualized in imputation_viz.[i]py[nb]

Incidentally closes #29


Snakefile
```diff
@@ -4,7 +4,7 @@ SUPERVISED, = glob_wildcards("model_configs/supervised/{supervised}.yml")
 IMPUTE, = glob_wildcards("model_configs/imputation/{impute}.yml")
 UNSUPERVISED, = glob_wildcards("model_configs/unsupervised/{unsupervised}.yml")
 SEMISUPERVISED, = glob_wildcards("model_configs/semi-supervised/{semisupervised}.yml")
-NUM_SEEDS = 5
+NUM_SEEDS = 3
```
jjc2718, line 7: Why did you decide to reduce the number of random seeds? Do you want this to apply to all the experiments (it looks like it will here), or just the imputation pretraining ones?

saged/imputation_pretraining.py
```diff
@@ -0,0 +1,303 @@
...
+                        help='The number of splits to use in cross-validation',
+                        type=int,
+                        default=5)
+    parser.add_argument('--batch_correction_method',
+                        help='The method to use to correct for batch effects',
+                        default=None)
+
+    args = parser.parse_args()
+
+    # TODO
```
jjc2718: Do you want to make this into an issue?

saged/imputation_pretraining.py
```diff
@@ -0,0 +1,303 @@
...
+    imputation_model.model = imputation_model.model.to('cpu')
+
+    imputation_models.append((imputation_model, 0))
+    subset_percents.append(0)
+
+    # Imputation training loop
+    for subset_number in [1, 10]:
+        subset_percent = subset_number * .1
+
+        # TODO fix this logic if you want to track impute performance
```
jjc2718, line 150: You could also consider making this an issue (although I don't feel too strongly either way, I know I leave tons of TODOs in my code)

saged/imputation_pretraining.py
```diff
@@ -0,0 +1,303 @@
+"""
+This benchmark trains an imputation model for downstream use in classification
```
jjc2718: This script also does the classification/evaluation, right? Might be good to note that here

saged/models.py
```diff
@@ -1406,3 +1506,79 @@ def fit(self, dataset: MixedDataset) -> "PseudolabelModel":
...
+                 train_count: float = None,
+                 **kwargs,):
+        """
+        Standard model init function for a supervised model
+
+        Arguments
+        ---------
+        pretrained_model: A model trained via imputing gene expression data
+        optimizer_name: The name of the optimizer class to be used when training the model
+        loss_name: The loss function class to use
```
jjc2718, line 1536: I think you're missing `model_name` here

notebook/converted/imputation_viz.py
```diff
@@ -123,3 +123,88 @@
...
+
+
+sepsis_df = metrics[metrics['disease'] == 'sepsis']
+tb_df = metrics[metrics['disease'] == 'tb']
+
+
+# In[19]:
+
+
+ggplot(sepsis_df, aes(x='train sample count', y='balanced_accuracy', color='factor(trial)')) + geom_line()
```
jjc2718, line 174: What is each factor here? Are they different imputation proportions, or different holdout studies/holdout sets?

Also, do you have any idea what's causing the oscillations here (accuracy goes up and down as train sample count increases) in some of the trials?

notebook/converted/imputation_viz.py
```diff
@@ -123,3 +123,88 @@
...
+
+
+# In[20]:
+
+
+plot = ggplot(sepsis_df, aes(x='train sample count', y='balanced_accuracy', color='factor(impute_samples)'))
```
jjc2718, line 180: Similar comment here, would be good to know exactly what the different factors/colors are.

jjc2718 (approved):
Looks good! Just a few small inline comments, nothing major.

It could be interesting to do some kind of comparison between the imputed and non-imputed models in the future (although I suspect you're already thinking about what to do next).

Snakefile
ben-heil: Based on the the previous experiments, it seemed like the lack of signal would be pretty clear even with fewer trials, so I reduced the number of seeds to test and took a few hours off the runtime

I'm fine with the seed change applying to all the experiments, though I may increase it in the future if I can ever get something that looks like a signal

saged/imputation_pretraining.py
ben-heil: Those actually correspond to things I've implemented so I'll go ahead and remove them

saged/imputation_pretraining.py
ben-heil: I think it makes more sense for it to live in this file for now since I'm pivoting the project and this file may not be used again for awhile

saged/imputation_pretraining.py
ben-heil: Good catch, updated the comment accordingly

saged/models.py
ben-heil: I actually forgot to remove it from the parameter list. It gets provided by the calling code (the original PytorchImpute instance has to know which model it's using), but TransferClassifier itself doesn't really care since it is given a pretrained model as a parameter

notebook/converted/imputation_viz.py
ben-heil: I called the datapoints corresponding to a series of models trained on increasing subsets of a single dataset split a "trial" for the sake of this notebook. I'm not really sure what causes the oscillations in accuracy, I'd guess that it's just noise caused by different training data

notebook/converted/imputation_viz.py
ben-heil: Makes sense, I added a note before the first visualization in the notebook since I use the term "trial" throughout without explaining it.

PR #35, opened 2021-05-06
Simulate Expression Data

ben-heil:
This PR adds the code required to use `ponyo` to simulate expression data. In the near future we will use the simulated data to verify that the model training code used in the rest of `saged` is working correctly


saged/simulate_expression.py
```diff
@@ -0,0 +1,257 @@
...
+    labeled_scaler.fit(train_data)
+
+    train_samples = set(disease_data.get_samples())
+    disease_only_data = copy.deepcopy(disease_data).subset_samples_to_labels([args.label])
+
+    data_copy = copy.deepcopy(disease_data)
+    healthy_only_data = data_copy.subset_samples_to_labels([args.negative_class])
+
+    # Determine the held out samples
+    held_out_samples = [sample for sample in all_disease_samples if sample not in train_samples]
```
jjc2718, line 109: It doesn't look like you're using these samples anywhere in this script - are you planning to use them for something later?

saged/simulate_expression.py
```diff
@@ -0,0 +1,257 @@
...
+
+    metadata['healthy_loss'] = hist.history['loss']
+    metadata['healthy_val_loss'] = hist.history['val_loss']
+
+    healthy_simulated = run_sample_simulation(healthy_encoder,
+                                              healthy_decoder,
+                                              healthy_df,
+                                              args.sample_count)
+
+    # Load all data not in healthy/disease
```
jjc2718, line 181: Maybe I'm forgetting from your past experiments with unlabeled data, but why would a data point not be in the healthy or disease sets? Are you only using certain studies to train/evaluate your model on?

notebook/debugging/visualize_simulation_data.ipynb
```diff
@@ -0,0 +1,1545 @@
...
...rCopOT63PR2lWfm7Lpo0cXVyQU5LpUeAmz2xmy5Yt9QJwn89HsdWKM9t+QRBoZ7fi3ckPQcxONutg8Pnw+33IRlNGqSiZwL6THrbBYKDYnUc4EiEcDKCm09gKC1FFEU80QlpRdtPPPtD44YcfeO2VVwiFQvQ/6ihGjhzJ+1OnMmnCBH7z+xl0+hnccdddvPzKK6xc8gsTjj6c4jwXP1VW8dhnn6NoGl//sZ0B5WWkVZWvt1fRql2H3P146913GXXVVVzy5Xx0Oh3nX3AB14waRSit1J8IyDLpbDZ+5OWX8+UXX9DaZmVFTS1HlZVg1MkMf+stnnjySTq2a8f0DVs4tKiAQCpNviuPib9twG0109piYdwvK2jTsmVOknZPmD17Npf9619c0rkDrcqKeGvBt/y2Zg0bN22id0Ee7Ww2XlrwHT//+CNvv/tubtI4ceJEfvjuO2wOB1eMHMlhhx3G8889x9BeXVE0lZ+qanAY9GyPRKmqqqK0tLQ5a9oIli5dyoIFC2jVqhV6vZ5kMsmCBQvo1asXvXv33qdjhcPh3RIee3Iunjt3Lv369SORSPDoo48yaNAgVq9ejd1uJxwO13u3CoKA3W7/23TNm/HPozkAB26/4w6++/ZbVv36K6Ig1KOC1GUzLrjggkb3P+SQQ/h67lw+/fRT3psyhcFWE8PatgKg3GalpcuJX4NnVqxm/BGHogHPLPuVcquF+39aynKPjydGj6bDpElUVVVhyepNNwXxeJzaUBiT08lJZ5yByW7n1eee5VevlxPOOJNTTzuNM047jbsWLuKQwnw+2bINRac/YJq3qVSK2kAQg92OMUubCHo9GCORjKGK00na46Fbhw7EEkkenb+A89u2YpbXz8cbNvPmfWN59MEHOam0CL0kkVAUKmxWzHodmzZtalIADpnBVhTFesvIdQGZpmkg1M+2SZKEmk7tdhxFURB1+ty2Op0OQZKbxJuXJAlJFFFVhVQiiWQykU6l0RSlSVJ9iUSCSDqNI7+gXmYuEok02h/MRiNJVSPoqSUWjaIkYqiJJGZZalLwXevxZNwoJZl4MsmmDRsoKimmpsZDNBHns/nf0qKoiIf7HIxTFjmxpJirl6xk4ZKltOvSFW8oiJTN1O8Ner2eAruNWr8fv9/P5uoqXO4CqpQ0oWCYViVFFJWWEovHqfR4cTocvPXWm7z37mT
```
jjc2718, line 807: It looks like there are some healthy samples that are outliers here in the UMAP plot, but the simulated data tends to be more similar to the "normal" healthy samples - I guess that's pretty typical for VAEs (they tend to underestimate/shrink data variance, I know Alex has come across this in her work) but just something to keep in mind as you move forward.

jjc2718 (approved):
Looks good to me!

How are you planning to use the simulated data? Are you going to use it similarly to the real data to estimate learning curves, etc., or is there some other way you can use it to validate your pipeline?

saged/simulate_expression.py
ben-heil: Right, one of the experiments that I hope to run is to see whether training with simulated + real data leads to better model performance, so I needed to hold out samples from VAE training

saged/simulate_expression.py
ben-heil: I'm using `disease_samples` as a general term, since the disease_samples won't necessarily correspond to sepsis. That being said, `disease_samples` isn't all labeled nonhealthy samples, just the ones with the label specified in `args.label`

PR #37, opened 2021-05-24
Evaluate Simulation

ben-heil:
Previous work on real expression data led to confusing results.  We found high-performing and low-performing subsets of data, weird reactions to batch effect correction, and so on. Because we believe that this weirdness is caused by uncontrolled inter-study/inter-platform heterogeneity, we decided to simulate data. If the methods worked on the simulated data, then we know that the previous issues we ran into were data issues, not code issues. 

This PR evaluates simulated data from the previous PR using the `keep_ratios` training script that was used to evaluate performance on real data. Logistic regression has no problem differentiating between simulated healthy and sepsis data, but the three layer neural network model has weird bimodal results based on how the datasets are split. We no longer have study structure to the data that could possibly skew these splits, and the label distribution between the train and validation sets should be identical. We also know the classification problem is a relatively easy one that is linearly separable based on the linear regression results. These negative results indicate that I need to take a closer look at what's going on in the training code, which will be a future PR.

image
image

For now, the changes:
- Updated the snakefile to run `keep_ratios` on the simulated data
- Created a notebook called `simulation_results` to visualize the results
- Wrote a script called `create_simulation_metadata` to generate the required files to run the outputs of `simulate_expression.py` through `keep_ratios.py`
- Minor bugfixes and code DRYing


ben-heil:
I'm not sure why coveralls is dying, but I'm planning on doing a PR to do conda environment caching soon so I'll fix it at the same time

saged/create_simulation_metadata.py
```diff
@@ -0,0 +1,74 @@
+""" This script creates the metatdata required for simulated data to be used in
```
jjc2718:
```suggestion
""" This script creates the metadata required for simulated data to be used in
```

saged/create_simulation_metadata.py
```diff
@@ -0,0 +1,74 @@
...
+
+        # Generate dummy experiment and sample ids for use in
+        sample_ids = ['{}_{}'.format(label, i) for i in range(sample_count)]
+        current_df.index = sample_ids
+        experiment_ids = ['Experiment_{}_{}'.format(label, i) for i in range(sample_count)]
+
+        label_to_samples[label] = sample_ids
+        labeled_samples.update(sample_ids)
+
+        # Write metadata in refinebio format for the samples
```
jjc2718, line 49: Do you have any documentation anywhere of what metadata is required to run your classifiers? That may be a good thing to give an example of somewhere (doesn't have to be part of this PR, but maybe in the future when you have things working well).

saged/create_simulation_metadata.py
```diff
@@ -0,0 +1,74 @@
...
+    all_data_df = all_data_df.transpose()
+
+    # Save compendium
+    all_data_df.to_pickle(args.compendium_out_file)
+
+    # Save label file
+    with open(args.label_out_file, 'wb') as out_file:
+        pickle.dump((label_to_samples, labeled_samples), out_file)
+
+    # Save metdata
```
jjc2718:
```suggestion
    # Save metadata
```

notebook/converted/simulation_results.py
```diff
@@ -0,0 +1,228 @@
...
+sepsis_metrics
+
+
+# In[9]:
+
+
+plot = ggplot(sepsis_metrics, aes(x='train_count', y='balanced_accuracy', color='supervised')) 
+plot += geom_smooth()
+plot += geom_point(alpha=.2)
+plot += geom_hline(yintercept=.5, linetype='dashed')
```
jjc2718, line 199: I assume this dotted line is some kind of baseline (maybe always predicting the most common class)? Interesting that the NN predictions seem to either be exactly random or close to 1, with only a few exceptions.

I agree with you that this probably points to a training/optimization issue, at least based on my experience with this kind of thing.

notebook/converted/simulation_results.py
```diff
@@ -0,0 +1,228 @@
...
+# In[2]:
+
+
+def split_sample_names(df_row):
+    train_samples = df_row['train samples'].split(',')
+    val_samples = df_row['val samples'].split(',')
+    
+    return train_samples, val_samples
+
+def create_dataset_stat_df(metrics_df, sample_to_study,
```
jjc2718, line 30: Could you put any of these functions into a file that's shared between analysis scripts? This code looks similar to what you have in some of your other notebooks (although maybe I'm missing some differences).

jjc2718 (approved):
Looks good - just a few small things.

saged/create_simulation_metadata.py
ben-heil: I really struggle to spell anything with the word data in it for some reason. I've actually thought about writing a macro to change all instances of "datset" in my code to "dataset" hahaha

saged/create_simulation_metadata.py
ben-heil: I don't, and I should. Added https://github.com/greenelab/saged/issues/38 to address this in the future

notebook/converted/simulation_results.py
ben-heil: Yeah, that's a good idea. I copied the logic over to generate results quickly, but it definitely needs DRYing

PR #39, opened 2021-06-10
Fixing Model Training

ben-heil:
Positive results in a positive control? What does that mean? Do I not have to entirely rework my experiment plans this time?

I simulated data in the previous PR, and found that my neural nets failed to learn even clearly linearly separable problems. I dug into the problem and found that instability in training was leading to the issues. One cause of this was very large gradients early on in training, which I fixed via gradient clipping. The deep neural net also had dropout before batchnorm instead of after, which caused more instability. With those problems fixed, I reran the models and found they were able to separate the data fairly easily.

Now that the models work I will rerun the models on real data to see what training looks like with models that work


ben-heil:
Conda is failing to install the environment, but mamba installs it just fine. I'm waiting the hour(s) it takes for conda to resolve the conflict, but I may just end up finally switching over to setup-miniconda so I can use mamba with Github Actions (that would be the next Pr though).

saged/models.py
```diff
@@ -1037,12 +1054,17 @@ def fit(self, dataset: LabeledDataset) -> "PytorchSupervised":
                 loss = self.loss_fn(output, labels)
 
                 loss.backward()
+
+                if self.clip_grads:
```
jjc2718, line 1058: Did you try the gradient clipping separately from the other changes (change in learning rate and dropout after batch norm)? Curious if you have a good idea about which change made the most difference, or if you need the combination of all of them.

I guess from your perspective it probably doesn't matter that much (if it works, it works), just wondering.

notebook/analysis/simulation_results.ipynb
```diff
@@ -16,103 +25,508 @@
...
+     },
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "plot = ggplot(sepsis_metrics[sepsis_metrics['supervised'] == 'three_layer_net'], aes(x='balanced_accuracy')) \n",
+    "plot += geom_density()\n",
+    "plot += ggtitle('Simulated Sepsis Accuracy vs Sample Count')\n",
```
jjc2718, line 456: This title shouldn't be "vs Sample Count", should it? Maybe something like "Distribution of simulated sepsis accuracy"?

notebook/analysis/simulation_results.ipynb
```diff
@@ -426,32 +837,32 @@
...
      },
-     "execution_count": 8,
+     "execution_count": 10,
      "metadata": {},
      "output_type": "execute_result"
     }
    ],
    "source": [
-    "plot = ggplot(sepsis_metrics, aes(x='train_count', y='balanced_accuracy', color='supervised')) \n",
+    "plot = ggplot(tb_metrics, aes(x='train_count', y='balanced_accuracy', color='supervised')) \n",
```
jjc2718, line 849: Why do you think the `three_layer_net` and `pytorch_lr` models require so many more samples than the other models? Do you think more parameter tuning/model tweaking would help with this?

jjc2718 (approved):
Nice work! Always good when the positive control actually turns out to be a positive control 😂

saged/models.py
ben-heil: I tried them separately, and the gradient clipping was by far the biggest difference for the three layer net. The deep net didn't care much about any of those changes, it just needed batchnorm to be before dropout

notebook/analysis/simulation_results.ipynb
ben-heil: Good catch!

notebook/analysis/simulation_results.ipynb
ben-heil: I'm really not sure. It feels like there's some flavor of local minimum that you can fall into and the regularization of 'this model just has a tooon of parameters' doesn't kick in until deep_net

PR #40, opened 2021-06-11
Added mamba workflow

ben-heil:


ben-heil:
Merging without review because workflows can't be tested until they're already in the repo

PR #41, opened 2021-06-11
Workflows

ben-heil:
This PR updates my workflows to use the `setup-miniconda` action and enables test coverage calculation. The tests run fine, but they fail to connect that information to the statements they covered for some reason. I've spent enough time tinkering with it so I'll move on for now unless you (or @ajlee21 ) know what might be causing the issue


ajlee21 (commented):
Will coveralls work with bash?

Here is my coveralls for ponyo running on pytest though: https://github.com/greenelab/ponyo/blob/master/.github/workflows/tests.yaml

ben-heil:
> Will coveralls work with bash?
> 
> Here is my coveralls for ponyo running on pytest though

Yeah it's running correctly, it's just not seeing the connection between the test code and the code it's testing for some reason. Also I think your link didn't send

jjc2718 (approved):
Changes look good to me!

I haven't used coveralls before so I don't have any ideas on what might be wrong with it - sorry. Fine with moving on since it doesn't seem critical, and the tests seem to pass otherwise.

PR #42, opened 2021-06-16
Disease vector

ben-heil:
As one final check to make sure everything works, I wrote a simulation script `disease_vector.py` that does the following
1. Trains a VAE on all sepsis and healthy data
2. Calculates the centroids of the sepsis and healthy embeddings
3. Does linear interpolation of centroids from the healthy to sepsis mean
4. Simulates data at different interpolated means to create classification tasks with varying difficulty

I then wrote `classify_disease_vector.py` to run the classification pipeline on the simulated data. The results are very nice (as in the models all train like you would expect, and the simulated data gets harder to predict as the distributions get closer together). As far as I can tell, everything in the pipeline is working, which means that the data is the issue in the original classification pipeline.


saged/disease_vector.py
```diff
@@ -0,0 +1,170 @@
...
+    parser.add_argument('compendium_metadata',
+                        help='A json file containing information compendium_file')
+    parser.add_argument('out_dir', default='data/simulated',
+                        help='The directory to store the simulated data files in')
+
+    parser.add_argument('--simulation_config', default='model_configs/simulation/tybalt.yml')
+    parser.add_argument('--sample_count', default=1000, type=int,
+                        help='The number of samples to generate for each translated dataset')
+
+    parser.add_argument('--label',
```
jjc2718, line 36: Should `args.label` and `args.negative_class` cover the whole dataset? Like in this case any of the input samples should be either sepsis or healthy (i.e. none of them are TB samples or any other non-healthy, non-sepsis condition)?

saged/disease_vector.py
```diff
@@ -0,0 +1,170 @@
...
+        full_sample[i, :] = np.random.normal(loc=disease_centroid,
+                                             scale=disease_std)
+
+    sim_disease = expression_decoder.predict_on_batch(full_sample)
+
+    out_path = os.path.join(args.out_dir, 'joint_sim_disease.pkl')
+    with open(out_path, 'wb') as out_file:
+        pickle.dump(sim_disease.T, out_file)
+
+    # TODO save result
```
jjc2718: Looks like you're doing this already, so maybe you can remove this TODO?

saged/disease_vector.py
```diff
@@ -0,0 +1,170 @@
...
+
+    disease_healthy_difference = disease_centroid - healthy_centroid
+    for interpolation_amount in range(0, 11, 2):
+        # Calculate the shifted centroid
+        sim_centroid = healthy_centroid + (interpolation_amount * .1 * disease_healthy_difference)
+
+        # Is there a good way to interpolate the standard deviation for each LV?
```
jjc2718, line 152: Have you looked at how much the healthy vs. diseased standard deviations vary in the latent space? In the UMAP plots the data variances look pretty similar, so maybe you don't need to interpolate at all.

saged/disease_vector.py
```diff
@@ -0,0 +1,170 @@
...
+        # Linear interpolation seems dicier for scale than for location, but that may just be me
+        sim_std = healthy_std
+
+        # Sample data from the new centroid
+        full_sample = np.zeros((args.sample_count, disease_embeddings.shape[1]))
+        for i in range(args.sample_count):
```
jjc2718, line 158: Can you use the `size` parameter to `np.random.normal` here instead of the for loop, since the mean and variance is the same for each sample? Something like this might work:
```python
full_sample = np.random.normal(loc=sim_centroid, scale=sim_std, size=(args.sample_count,))
```

jjc2718 (approved):
Looks good to me!

A few questions on `simulate_results.ipynb` (diff is too big to comment inline):

* Are the UMAP plots of the generated data in the original feature space? Or in the latent space? Would the plots look any different between the two?

* In the facet plots at the end (comparing accuracy and amount of data used for different interpolation ratios), the confidence intervals for the `deep_net` model are pretty large compared to logistic regression. Are you worried about this, or do you think it'll be less of a problem on larger datasets?

saged/disease_vector.py
ben-heil: They should cover the entire dataset, but since the arrays themselves aren't labeled we need to pass in the information so that the script knows what disease it's working with

saged/disease_vector.py
ben-heil: Good catch

saged/disease_vector.py
ben-heil: I agree, the simulation looks pretty good. The standard deviations do differ between the classes, but apparently not enough to worry about. I think I'll stick with not interpolating

saged/disease_vector.py
ben-heil: That feels like it should work, but I wasn't able to get it to cooperate when I tried. Since this script is only generating a few thousand samples I figure the compute gains from avoiding a loop won't match the programmer time lost by fixing it :)

PR #43, opened 2021-07-13
Performer Self Attention

ben-heil:
This PR adds a model that imputes expression data using a linear runtime self-attention layer (implementation from https://github.com/lucidrains/performer-pytorch).   

Results so far are that it seems to work, but it explodes when it sees real datasets. Also the layernorm seems to hurt instead of help. My guess is that it's a data normalization issue, so I'm going to come back to it once I have the recount3 data ready

Changes:
- The model implementation is in `models.py`, the other changes are config files and minor changes to make it possible to run/test the new model


notebook/debugging/create_smol_dataset.ipynb
```diff
@@ -0,0 +1,566 @@
+{
```
jjc2718, line 1: Is this dataset just for making sure everything works? It might be better to randomly sample 20 rows instead of just taking the top 20 (for accurate representation across studies, etc), unless you're sure it doesn't matter.

saged/models.py
```diff
@@ -383,6 +384,89 @@ def set_final_layer(self, new_layer: nn.Module):
...
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        # Duplicate the data along the "embedding dimension"
+        x = x.unsqueeze(2).repeat(1, 1, self.dim)
+        for i in range(len(self.layers)):
+            x = self.layers[i](x)
+
+            # Don't apply dropout to the final layer
+            if i < len(self.layers) - 1:
+                x = self.dropout(x)
```
jjc2718: So you do want to apply dropout to the performer blocks? Is that the standard approach?

model_configs/imputation/three_layer_performer.yml
```diff
@@ -0,0 +1,16 @@
+name: "PytorchImpute"
+optimizer_name: "Adam"
+loss_name: "MSELoss"
+model_name: "ImputePerformer"
+layer_count: 3
+num_heads: 8
+dim: 8
+lr: .001
```
jjc2718, line 8: Do you think these hyperparameters will need to be tuned? Or are these models generally fairly robust to hyperparameter settings?

saged/impute_expression.py
```diff
@@ -67,7 +61,13 @@
     subset_percents = []
     val_losses = []
     for i in range(len(cv_splits)):
-        for subset_number in range(1, 11):
+        for subset_number in [2, 4, 8]:
```
jjc2718, line 64: Is there a reason why you don't want to just loop through `subset_percent` here directly, rather than looping through `subset_number` then converting it to `subset_percent` on line 71? It doesn't look like you're using `subset_number` anywhere else, but maybe I'm missing something.

jjc2718 (approved):
Looks good! A few small comments.

I think you're starting to outpace my deep learning knowledge a bit at this point, so maybe it would be good to have David take a look at some of your PRs too going forward (I'm guessing he's probably the one in the lab who knows the most about this stuff other than you).

I'm happy to keep reviewing too, of course - just realizing as I frantically Google "performer deep learning" that I may miss some basic things 😂

saged/models.py
ben-heil: Maybe? So I knew that the implementation I was working from [used dropout layers](https://github.com/lucidrains/performer-pytorch/blob/main/performer_pytorch/performer_pytorch.py#L373), but I thought it used them in the model portion rather than after the attention blocks. (Side note, the attention blocks are attention + one FC layer, which is why there's dropout.)

Anyway, if I decide to use dropout, I should do it through the dropout parameter already present in the attention block. And since it's set to zero by default, I'll lean towards not using it for now. Great catch!

notebook/debugging/create_smol_dataset.ipynb
ben-heil: Casey said the exact same thing to me in our meeting yesterday hahaha. It probably matters, but `smol_dataset` is mainly for finding runtime errors quickly. I should change it, but it's not too high on the list of priorities yet

model_configs/imputation/three_layer_performer.yml
ben-heil: They'll probably need to be tuned some day, but I'm putting that off until I have a working model and dataset :)

saged/impute_expression.py
ben-heil: Yeah it's an artifact of the old `in range(1, 11)` line here. Range doesn't let me iterate by .1 (or at least I wasn't able to figure out how to force it to), so I stuck with the 1-10 range. I imagine I'll switch back to a linear progression at some point so I'm fine with having this be two statements for now

PR #44, opened 2021-07-16
Download and normalize recount3 data

ben-heil:
This PR contains the scripts required to download the recount3 data and metadata. 

I think the main things to check are:
1) Is the [online variance calculation](https://github.com/greenelab/saged/compare/master...ben-heil:recount3_data?expand=1#diff-cc0c95cf7a6b4cfc39bd815b80a136b61752c77eac2877dd7187f07bfbb1a505R112-R123) in `normalize_recount.py` implemented correctly
2) Is my R style correct and my code efficient enough to rank somewhere above "absolutely horrible"


saged/download_recount3.R
```diff
@@ -0,0 +1,239 @@
...
+#install.packages("tidyr", repos = "https://cran.r-project.org", dependencies=TRUE)
+#BiocManager::install()
+#BiocManager::install("recount3")
+
+library(recount3)
+library(dplyr)
+library(tibble)
+library(tidyr)
+
+`%>%` <- dplyr::`%>%`
```
arielah: I don't think you need this? %>% inherits from magrittr in both dplyr and tidyr, so it shouldn't change. Let me know if I'm wrong about that.

saged/download_recount3.R
```diff
@@ -0,0 +1,239 @@
...
+
+getMetadata <- function(project, project_home, metadata_df) {
+  # Append the metadata for a project to a dataframe
```
arielah, line 26: You might not want to do this given you're only doing this part of the project in R, but have you considered using the `SummarizedExperiment` class from Bioconductor? I've found it the most painless way to associate data with metadata in R.

saged/download_recount3.R
```diff
@@ -0,0 +1,239 @@
...
+    )
+  }
+  if (is.null(result)) {
+    return(metadata_df)
+  }
+  return(result)
+}
+
+getCounts <- function(project, project_home) {
+  # Append the metadata for a project to a dataframe
```
arielah: This is the same description as the `getMetadata` function, probably a copy-paste mistake

saged/download_recount3.R
```diff
@@ -0,0 +1,239 @@
...
+        return(NULL)
+      }
+    )
+  }
+  return(counts)
+}
+
+
+# Make sure R's working directory is in the correct spot to make relative paths resolve correctly
+if (rstudioapi::isAvailable()) {
```
arielah, line 126: Take it or leave it, but what I do is have a config file with the paths where I want data to be accessed from, e.g. `data_dir <- normalizePath("~/folder/of/interest")` and call `source(config.R)` and direct all paths relative to data_dir. Idk if that's helpful, but calling the rstudioapi scares me haha

saged/download_recount3.R
```diff
@@ -0,0 +1,239 @@
...
+  # Get only sra data
+  if (home != "data_sources/sra") {
+    print(paste('Skipping ', project, 'from', home))
+    next
+  }
+  
+  print(project)
+  metadata_df <- getMetadata(project, home, metadata_df)
+  
+  if (i %% 10 == 0) {
```
arielah, line 184: So you're saving the mega-file every 10th run through in case of failure? It is easy to tell where the script failed so you can start appending from that point?

saged/download_recount3.R
```diff
@@ -0,0 +1,239 @@
...
+  # As a result, we need samples to be rows, not columns
+  counts <- t(counts)
+  
+  # write header with data
+  if (i == 1){
+    write.table(counts, file=out_file, col.names=TRUE, sep='\t', )
+  }
+  
+  # Write counts to file
+  write.table(counts, file=out_file, append=TRUE, sep='\t', col.names=FALSE)
```
arielah: If it is the first project (i==1), wouldn't the data then be written twice? It seems to me like this should be an if-else statement

saged/get_gene_lengths.R
```diff
@@ -0,0 +1,45 @@
...
+#install.packages("tidyr", repos = "https://cran.r-project.org", dependencies=TRUE)
+#BiocManager::install()
+#BiocManager::install("recount3")
+#BiocManager::install("EDASeq")
+
+library(recount3)
+library(dplyr)
+library(tibble)
+library(tidyr)
+library (EDASeq)
```
arielah: TIL you can have a space between library and ( and it doesn't ruin everything. Cool!

saged/get_gene_lengths.R
```diff
@@ -0,0 +1,45 @@
...
+  return (unlist(strsplit(name, '\\.'))[[1]])
+}
+
+# The project doesn't really matter, I'm just using it to get the genes used in recount3
+url <- recount3::locate_url(project='SRP103067', project_home='data_sources/sra', type='gene')
+count_files <- recount3::file_retrieve(url)
+counts <- recount3::read_counts(count_files)
+
+genes <- rownames(counts)
+genes_out <- unlist(lapply(genes, split_gene_name))
```
arielah, line 40: Pretty trivial, but if you want a one liner way to do this it's `genes_out <- gsub("\\..*", "", genes)`

saged/metadata_to_tsv.R
```diff
@@ -0,0 +1,25 @@
+# This script writes the metadata generated by download_recount3.R to a tsv
```
arielah, line 1: I'm confused. I'm guessing you're doing this because the metadata is big enough that it's unwieldy to write it out as a tsv in `download_recount3.R`, but if it's that big, why are you writing it out as an R object bunch of times, wouldn't that be slow/unwieldy too?

saged/normalize_recount_data.py
```diff
@@ -0,0 +1,178 @@
...
+            try:
+                gene_to_len[gene] = int(length)
+            except ValueError:
+                # Some genes have no length, but will be removed in a later step
+                pass
+    return gene_to_len
+
+
+def calculate_tpm(counts: np.ndarray, gene_length_arr: np.ndarray) -> np.ndarray:
+    """"Given an array of counts, calculate the transcripts per kilobase million
```
arielah, line 37: I thought TPM just meant "transcripts per million" without considering length, not "transcripts per kilobase million." Am I wrong?

saged/normalize_recount_data.py
```diff
@@ -0,0 +1,178 @@
...
+
+        gene_length_arr = np.array(gene_length_arr)
+
+        means = None
+        M2 = None
+        maximums = None
+        minimums = None
+
+        # First time through the data, calculate statistics
+        for i, line in tqdm.tqdm(enumerate(count_file), total=317475):
```
arielah: What is this `total=317475` number? Is that liable to change if recount changes at all?

saged/normalize_recount_data.py
```diff
@@ -0,0 +1,178 @@
...
+                # Online variance calculation https://stackoverflow.com/a/15638726/10930590
+                if means is None:
+                    means = tpm
+                    M2 = 0
+                    maximums = tpm
+                    minimums = tpm
+                else:
+                    delta = tpm - means
+                    means = means + delta / (i + 1)
+                    M2 = M2 + delta * (tpm - means)
```
arielah, line 127: Tiny tiny thing, but why not say `M2 = M2 + delta * delta`?

arielah (approved):
Looks good! The online variance calculation is clever and looks correct to me. Only a few surface-level things to recommend.

saged/download_recount3.R
ben-heil: That's a good idea! For my use case I actually prefer the data and metadata to be separate from each other since I'll load them at different times. I think I'm fine with writing them to tsvs to be loaded in python, but I'll definitely remember `SummarizedExperiments` for future projects

saged/download_recount3.R
ben-heil: Good catch!

saged/download_recount3.R
ben-heil: Right, the metadata takes days to download, so I wanted to save fairly frequently. This way not too much is lost if the program crashes or I restart my computer

saged/download_recount3.R
ben-heil: :facepalm: Yep, you are correct

saged/metadata_to_tsv.R
ben-heil: I'm trying to make the data downloading code as independent of the rest of my pipeline as possible. Afaict no one has put together a script that downloads all of recount3 yet, so I will hopefully give the code to the recount3 people or publish it somewhere. If I do, I don't want future scientists to be bound to tsvs just because that's what I was using in my pipeline.

All of which is to say that you're correct, it would make more sense to write out the tsv in `download_recount3`, but I don't currently plan on doing so

saged/get_gene_lengths.R
ben-heil: Yep, definitely me intentionally exploring the boundaries of R whitespace :)

saged/normalize_recount_data.py
ben-heil: That would make sense given the acronym. It seems to be [used interchangeably](https://www.google.com/search?q=transcripts+per+million+vs+transcripts+per+kilobase+million) on the internet, so I'll add a link to the definition I'm working from in the function comments

saged/normalize_recount_data.py
ben-heil: Good question, I thought the same thing when I first implemented the calculation. Since `means` is redefined on 119, the values for `delta` and `tpm - means` on line 120 are different

saged/normalize_recount_data.py
ben-heil: 317475 is the number of lines in the file. Tqdm generates a progress bar for the loop and estimates the time remaining. Since reading all the way through the file to calculate length is slow, I just gave it the number of lines manually. This number will change with recount3, but the bug that occurs will be a slight misestimation of how long the loop will take.

To avoid using magic numbers I'll add a constant explaining what that number is

PR #46, opened 2021-08-02
Recount3 data

ben-heil:
A positive result on a proof of concept experiment? This is unprecedented in my career in the GreeneLab :cake: 

More seriously, instead of predicting diseases as I have in the past, this PR predicts tissue type from gene expression. More specifically, the results visualized in the new section of `benchmark_results` differentiate between breast tissue and blood tissue, the two most common tissue types in the manually labeled set in recount3.

Changes:
- Updated Snakefile to run the experiment on the new data
- Visualized the findings in `benchmark_results` (sorry about the way the converted notebook renders as being entirely new now that I moved it. Only the last 40 lines are new)
- Updated `normalize_recount_data` to remove duplicate samples
- Updated `pickle_tsv` so it can run on files that don't fit into memory
- Added `predict_tissue.py`, but it's mostly just a tweaked version of `keep_ratios.py` so doesn't need to be looked at too closely. At some point I should figure out how to refactor it and its siblings to move the shared code into a util file
- Added functions to parse recount data to `utils.py`


saged/impute_expression.py
```diff
@@ -35,16 +35,12 @@
 
     args = parser.parse_args()
 
-    # TODO
-    # Set early stopping criteria
-    # Evaluate effects of study splits (am I correct to split by study to avoid
-    # leakage?)
-    # Implement adding noise?
+    # with open(args.dataset_config) as data_file:
```
jjc2718, line 38: Do you want to keep these commented out lines?

saged/normalize_recount_data.py
```diff
@@ -99,10 +103,23 @@ def calculate_tpm(counts: np.ndarray, gene_length_arr: np.ndarray) -> np.ndarray
...
 
+        samples_seen = set()
+
         # First time through the data, calculate statistics
         for i, line in tqdm.tqdm(enumerate(count_file), total=LINES_IN_FILE):
             line = line.replace('"', '')
             line = line.strip().split('\t')
+            sample = line[0]
+
+            # Remove duplicates
```
jjc2718, line 114: Will these be exact duplicate samples? So it doesn't matter if you take the first one or any of the subsequent ones? (It looks like here you're keeping the first one and skipping all the rest, just making sure that's what you want)

saged/pickle_tsv.py
```diff
@@ -12,7 +15,22 @@
...
+        for col in header:
+            dtypes[col] = np.half  # half precision floats
+
+    chunksize = 10000
+
+    df = pd.DataFrame()
+    data_iterator = pd.read_csv(args.tsv_path, sep='\t', index_col=0, dtype=dtypes,
+                                chunksize=chunksize)
+    for sub_data in tqdm.tqdm(data_iterator, total=32):
+        df = df.append(sub_data)
```
jjc2718, line 33: So eventually you will have all the data in memory, since you're creating one large dataframe, right? Is this possible because you're using half-precision floats (and the original dataset uses higher precision)?

saged/get_tissue_data.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import Dict
+
+import pandas as pd
+
+
+def map_samples_to_labels(file_path: str) -> Dict[str, str]:
+    metadata_df = pd.read_csv(file_path, delimiter='\t')
+    columns_to_keep = ['external_id', 'recount_pred.curated.tissue']
+
+    samples_and_labels = metadata_df.drop(metadata_df.columns.difference(columns_to_keep), 1)
```
jjc2718, line 17: If you want, you could try doing this using method chaining:
```python
samples_and_labels = (metadata_df
  .drop(metadata_df.columns.difference(columns_to_keep), 1)
  .dropna(axis='rows')
  .set_index('external_id')
)
```
I find it a bit easier to follow what's being done to the dataframe in each step with chaining, but definitely a personal preference thing - up to you.

saged/predict_tissue.py
```diff
@@ -0,0 +1,172 @@
...
+                        help='The yaml formatted dataset configuration file. For more information '
+                             'about this file read the comments in the example_dataset.yml file')
+    parser.add_argument('supervised_config',
+                        help='The yaml formatted model configuration file. For more information '
+                             'about this file read the comments in the example_model.yml file')
+    parser.add_argument('out_file',
+                        help='The file to save the results to')
+    parser.add_argument('--tissue1',
+                        help='The first tissue to be predicted from the data',
+                        default='Blood')
```
jjc2718, line 25: Do you have a list of the tissue options somewhere? You could use that here as the `choices` argument if you have it.

notebook/converted/benchmark_results.py
```diff
@@ -261,11 +310,818 @@
...
+
+
+# In[8]:
+
+
+plot = ggplot(tissue_metrics, aes(x='train_count', y='balanced_accuracy', color='supervised')) 
+plot += geom_smooth()
+plot += geom_point(alpha=.2)
+plot += geom_hline(yintercept=.5, linetype='dashed')
+plot += ggtitle('Brain vs Breast Tissue Prediction')
```
jjc2718, line 1125: Is this brain vs. breast or blood vs. breast? Seems like the filenames above say blood, but this title says brain.

notebook/converted/benchmark_results.py
```diff
@@ -261,11 +310,818 @@
...
+tissue_metrics['train_count'] = tissue_metrics['train sample count']
+
+tissue_metrics['supervised'] = tissue_metrics['supervised'].str.replace('pytorch_supervised', 'three_layer_net')
+tissue_metrics
+
+
+# In[8]:
+
+
+plot = ggplot(tissue_metrics, aes(x='train_count', y='balanced_accuracy', color='supervised'))
```
jjc2718, line 1121: Wow, interesting to see that the NNs provide such a big performance improvement here over the linear methods! Do you think this has something to do with the data (e.g. larger sample size, better label balance), or is there some biological reason that capturing non-linear relationships would help with tissue prediction?

jjc2718 (approved):
Looks good! Always nice to get the result you expect 🎉

saged/impute_expression.py
ben-heil: Probably not. I don't think I'll be going down the imputation route, so more compact code with less future ideas is better in `impute_expression`

saged/normalize_recount_data.py
ben-heil: They *should* be. I expect there are two main sources of duplication - stopping and restarting the download process, and samples being included in different studies. The samples from the former will be identical and the samples from the latter should still be identical (but may be slightly different due to preprocessing?). Either way, I don't want samples showing up multiple times within a study and having undue weight, and I definitely don't want the same sample showing up across studies and leaking information

saged/pickle_tsv.py
ben-heil: Correct, the original dataset is an ASCII flat file, so each entry has a ~10 bytes representing each number in the table. The final pickled dataframe ends up only being 3GB, but I added the changes here because I was running out of memory in initial runs. 

As it turns out, moving from 60k genes to 5k genes made the data smaller, but I left this file in its more general version in case I get bigger data

saged/get_tissue_data.py
ben-heil: Definitely will try to use method chaining more in the future, it's just not how my brain works by default unfortunately

notebook/converted/benchmark_results.py
ben-heil: My guess is that the strongest linear signals are batch effects within studies, and that the neural nets can pick up more nuance than that. I'm a little skeptical of the results though, I wouldn't expect a deeper network to strictly outperform the shallower one when only a few hundred samples are present. It may be a result of the deeper network being more regularized because it uses dropout and batch normalization :shrug:

notebook/converted/benchmark_results.py
ben-heil: Good catch!

saged/predict_tissue.py
ben-heil: Ooh good idea, will add it

ben-heil:
I'm going to add the two requested changes to the pull request I'm currently putting together to avoid potential merge conflicts

saged/normalize_recount_data.py
jjc2718: Got it, that makes sense! One thing you could do is to log the duplicates to a file somewhere to follow up on later, but if you're confident that you only want one maybe that's not necessary.

PR #47, opened 2021-08-04
Multi tissue

ben-heil:
This PR trains models to perform a multitask prediction task where a given sample of gene expression is classified into one of 21 tissue types. 

Changes
- Added the `choices` field in `predict_tissues` and fixed a title in `benchmark_results.ipynb` as requested in https://github.com/greenelab/saged/pull/46
- Updated the Snakefile to run `predict_tissues` with the new `--all-tissue` flag
- Added the `--all-tissue` flag to `predict_tissues`. This flag runs a multitask prediction for all labels with at least ten studies in the dataset

### DIFF
--- a/notebook/converted/benchmark_results.py
+++ b/notebook/converted/benchmark_results.py
@@ -1092,7 +1092,7 @@ in_files = glob.glob('../../results/Blood.Breast.*.tsv')
 print(in_files[:5])
 
 
-# In[6]:
+# In[3]:
 
 
 tissue_metrics = pd.DataFrame()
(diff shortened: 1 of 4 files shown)


notebook/converted/benchmark_results.py
```diff
@@ -1115,13 +1115,61 @@ def return_unlabeled():
...
+tissue_metrics['train_count'] = tissue_metrics['train sample count']
+
+tissue_metrics['supervised'] = tissue_metrics['supervised'].str.replace('pytorch_supervised', 'three_layer_net')
+tissue_metrics
+
+
+# In[9]:
+
+
+plot = ggplot(tissue_metrics, aes(x='train_count', y='balanced_accuracy', color='supervised'))
```
jjc2718, line 1164: I think you've done similar things in past experiments and I'm probably forgetting, but what is the classification task here? Are you doing something like one-vs-rest classification for each tissue type individually, or are you doing multiclass classification (without binarizing) and calculating accuracy over all classes?

My guess is the latter (since accuracy scores are lower overall than for the binary tissue classification task), but I just want to make sure my understanding is correct.

jjc2718 (approved):
Looks good! Are you planning to dig further into which tissues are easier or harder to predict (e.g. by looking at a confusion matrix or something similar), or is this problem more of a quick proof of concept to see that your models are learning correctly?

notebook/converted/benchmark_results.py
ben-heil: This is multiclass classification across all classes, and the metric is balanced accuracy (the mean of the recall of all classes). That is to say, you are correct :)

ben-heil:
> Looks good! Are you planning to dig further into which tissues are easier or harder to predict (e.g. by looking at a confusion matrix or something similar), or is this problem more of a quick proof of concept to see that your models are learning correctly?

I think plotting a confusion matrix will be useful at some point. For now I'm just looking at the relative difficulty of different prediction tasks and making sure models work

PR #48, opened 2021-08-06
Batch effect correction and deeper models

ben-heil:
This notebook examines the effects of batch effect correcting the data and running deeper models.

Results:
- LIMMA removes all the linear signals that are batch dependent. Because the tissue and study are heavily confounded, this removes all the biological signal models like LR would be able to detect. The nonlinear models can still find something, but performance is much worse
- Deeper models perform less well than the five layer neural network. I'd suspect this is due to vanishing gradients since I haven't added skip-connections or anything. Looks like the five layer network is the sweet spot for this dataset, and I'll avoid anything larger because those models take 5ever to train.

BE corrected binary classification:
image

More models on multiclass:
image

Changes: 
- Wrote new config files for the new models
- Visualized the results in the `benchmark_results` notebook
- Implemented a new class for arbitrarily deep fairly rectangular neural networks in `models.py`
- Modified `predict_tissues` to correct batch effects only on the labeled data because limma/combat wanted ~100s of GBs to batch effect correct the whole dataset

### DIFF
--- a/notebook/converted/benchmark_results.py
+++ b/notebook/converted/benchmark_results.py
@@ -1089,6 +1089,7 @@ plot
 
 
 in_files = glob.glob('../../results/Blood.Breast.*.tsv')
+in_files = [f for f in in_files if 'be_corrected' not in f]
 print(in_files[:5])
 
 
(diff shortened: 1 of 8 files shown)


saged/models.py
```diff
@@ -467,6 +467,58 @@ def set_final_layer(self, new_layer: nn.Module):
...
+
+        self.output = nn.Linear(input_size // 2, output_size)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = self.l1(x)
+        for layer in self.intermediate:
+            x = layer(x)
+        x = self.output(x)
+
+        return x
```
jjc2718, line 519: Does the `DeepClassifier` class (below) still have a use? It looks like it's essentially the same as this more general class, but with the number of layers hard-coded instead of passed in as a parameter.

notebook/converted/benchmark_results.py
```diff
@@ -1158,9 +1160,54 @@ def return_unlabeled():
...
+
+tissue_metrics['supervised'] = tissue_metrics['supervised'].str.replace('pytorch_supervised', 'three_layer_net')
+tissue_metrics['supervised'] = tissue_metrics['supervised'].str.replace('deep_net', 'five_layer_net')
+tissue_metrics
+
+
+# In[10]:
+
+
 plot = ggplot(tissue_metrics, aes(x='train_count', y='balanced_accuracy', color='supervised'))
```
jjc2718, line 1211: It's definitely interesting that the three-layer and seven-layer networks perform so much worse than the five-layer net. Maybe the learning rate or other hyperparameters could be affecting this? I guess it makes sense to avoid deeper networks if you're fairly sure it won't be an easy fix.

jjc2718 (approved):
Looks good to me!

saged/models.py
ben-heil: Yeah, the five layer network uses the DeepClassifier class as a base

notebook/converted/benchmark_results.py
ben-heil: Yeah for sure. I think it's some combination of being a good depth for the problem space and not running into vanishing gradients.

saged/models.py
jjc2718: Got it! Might be good to have them use the same base class in the future (that way if you make changes to one you won't have to remember to also make changes to the other), but I'm fine with it for now.

PR #50, opened 2021-08-19
Imputation pretraining (and a mystery!)

ben-heil:
Initially this was going to be a quick PR that showed that imputation pretraining did or didn't work for tissue prediction problems.

What ended up happening was these plots:
image
image

These results indicate that pretraining doesn't help (sure, that's not super surprising), but also that a three layer and five layer network perform the same (which contradicts the results from the last PR). In fact, all the models shown here, regardless of pretraining or architecture perform almost identical to the five layer network without imputation that was used in the last PR (see figure below )
image

I'm almost certain that this is a bug of some sort, but I haven't been able to figure out what it is. I tried making sure the architecture for the three layer net is exactly the same between the non-pretrained imputation model and the version used in the supervised PR. The code for the supervised portion of the training is almost identical, so there shouldn't be a difference occurring during training. It feels like the wrong model architecture is getting used somehow, but I have no idea how that might be happening.

Let me know if you see anything, the place where the mistake is most likely to be are in the main section of `imputation_pretraining.py` (see `predict_tissue.py` for the training logic that generated last week's figure).


saged/models.py
```diff
@@ -348,6 +379,65 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         return x
 
 
+class FiveLayerImputation(nn.Module):
+    """An imputation model based off the FiveLayerImputation model"""
```
```suggestion
    """An imputation model based off the DeepClassifier (five-layer) model"""
```
I think this is what you meant, since this class is `FiveLayerImputation`

model_configs/imputation/three_layer_impute.yml
```diff
@@ -6,7 +6,7 @@ lr: .001
 weight_decay: 0
```
jjc2718, line 6: Do you also want to have a config in `model_configs/imputation` for the five-layer imputation network? I don't see one in that directory currently (but maybe you just forgot to track it).

notebook/converted/benchmark_results.py
```diff
@@ -1215,6 +1215,57 @@ def return_unlabeled():
...
+plot += geom_point(alpha=.2)
+plot += ggtitle('Effects of Imputation on Multiclass Tissue Prediction')
+plot += facet_grid('impute_samples ~ .')
+plot
+
+
+# In[5]:
+
+
+plot = ggplot(tissue_metrics, aes(x='train_count', y='balanced_accuracy', color='factor(impute_samples)'))
```
jjc2718, line 1262: What is `factor(impute_samples)` here? My guess is that it's no imputation with `impute_samples == 0` vs. imputation with `impute_samples == 28436` - if that's right, it might be good to edit the legend to say that.

jjc2718 (approved):
Your code changes look good to me! Like you mentioned, the changes I can see to your Python scripts here are pretty minimal, which makes me think the problem may be somewhere in your Snakemake/results generation pipeline rather than in the added scripts that I can see here.

The only thing that occurs to me is the config file thing I pointed out - maybe somehow what you're plotting as the `five_layer` results is really the `three_layer` results (although I guess they're not identical, just statistically equivalent). If you do have a config file for the five layer experiments and you're just not tracking it in git, then I'm not sure what the issue might be.

Can you serialize/save the model architectures that are being loaded and look at them somehow? That might be a good debugging first step, to verify that you're using different architectures.

saged/models.py
ben-heil: Yep, good catch!

model_configs/imputation/three_layer_impute.yml
ben-heil: Right, I thought I had added it but apparently not

PR #51, opened 2021-08-31
Biobert

ben-heil:
Sorry, this is a big PR because the pieces ended up being interdependent.

Three main parts:
1) Implementing early stopping
2) Getting the imputation pipeline fixed
3) Writing the code for using biobert embeddings in addition to expression

The main things to review are:
1) Nothing in biobert-pytorch, that's just a copy of the bioBERT embedding repo
2) benchmark_results.yml - Source of the results shown below as well as rerunning analyses with the new early stopping regularization
3) extract_metadata_text - Pulls out metadata columns to feed into bioBERT
4)  models.py - early stopping code and refactored TransferClassifier to inherit from PytorchSupervised to avoid duplicating code (and fixing imputation in the process I think)
5) predict_tissue.py - added the logic to include the biobert embeddings in the training data
6) utils - Added a function to parse the h5py encoded bioBERT embeddings

New results:
Imputation works now and it's gorgeous. Five layer models are worse than three layer models with small amounts of data, but when pretraining via imputation five layer models outperform even at lower data settings
image


biobert-pytorch/LICENSE
```diff
@@ -0,0 +1,216 @@
+Disclaimer:
```
jjc2718, line 1: Do you need to track the `biobert-pytorch` directory in git? Could you just add it to your `.gitignore` file and provide instructions for downloading the code in your README?

Correct me if I'm wrong, but it doesn't seem like you're planning to modify the bioBERT code at all, just to use it to generate pretrained embeddings, so maybe you don't need to track the code.

saged/extract_metadata_text.py
```diff
@@ -0,0 +1,39 @@
...
+        for _, row in tqdm.tqdm(df.iterrows(), total=len(df)):
+            title = row['sra.sample_title']
+            experiment_description = row['sra.design_description']
+            sample_description = row['sra.sample_description']
+
+            metadata_string = ' '.join([str(title),
+                                        str(experiment_description),
+                                        str(sample_description)])
+
+            metadata_string = metadata_string.replace('nan', '')
```
jjc2718: Are you sure the string `'nan'` won't appear anywhere in the metadata other than actual NaN entries? It's only three letters, so it seems somewhat likely to occur by chance somewhere in normal text.

saged/predict_tissue.py
```diff
@@ -53,14 +56,50 @@
...
+                # Not all samples with metadata are in compendium
+                if sample not in expression_df.columns:
+                    continue
+
+                index_in_df = expression_df.columns.get_loc(sample)
+                placeholder_array[:, index_in_df] = embeddings[line_number, :]
+
+            # 0-1 normalize embeddings to match scale of expression
+            pa = placeholder_array
+            pa_positive = pa - np.min(pa, axis=0)
```
jjc2718: Could you use [MinMaxScaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MinMaxScaler.html) to do this scaling?

saged/predict_tissue.py
```diff
@@ -150,6 +189,26 @@
...
+                    save_path = os.path.splitext(save_path)[0]
+
+                    if args.all_tissue and args.biobert:
+                        extra_info = 'all_tissue_biobert'
+                    elif args.all_tissue:
+                        extra_info = 'all_tissue'
+                    elif args.biobert:
+                        extra_info = 'biobert'
+                    else:
+                        extra_info = '{}-{}'.format(args.tissue1, args.tissue2)
```
jjc2718, line 204: Are you planning to use bioBERT for these individual tissue comparisons too?

saged/utils.py
```diff
@@ -437,14 +438,48 @@ def load_recount_data(config_file: str) -> Tuple[pd.DataFrame, Dict[str, str], D
...
+    Read the biobert embeddings of the metadata for each sample in the dataset
+
+    Arguments
+    ---------
+    config_file: The file containing paths to the embedding file
+
+    Returns
+    -------
+    biobert_embeddings - A numpy array containing the biobert embeddings of the metadata in the
+                         same order as expression_df
```
jjc2718:
```suggestion
                         same sample order as expression_df
```
I think this is accurate? Since you want to generate different features to use in addition to the genes that you already have, but they should cover the same samples in the same order?

jjc2718 (approved):
Nice work! Your results definitely look promising. A few big-picture questions, in addition to the inline comments:

* Can you briefly explain what you're doing with the bioBERT embedding? It looks like you're extracting features from the text metadata for each sample, then using the resulting features in addition to gene expression to predict tissue?
* What's included in the metadata you're using? Are there particular parts of the metadata that you think are helpful for this task?

biobert-pytorch/LICENSE
ben-heil: I don't planning on modifying the bioBERT code (any more than I already have), but I also want to make sure that people using Saged are using the same version of the code that I used. If it were a module I could just pin the version when installing it, but it's not.

Also I'm only including part of the bioBERT code in the repo. I think this is the cleanest way to handle the dependency out of a number of ugly options

saged/extract_metadata_text.py
ben-heil: It doesn't, but I agree that I should fix that in case I get more metadata later

biobert-pytorch/LICENSE
jjc2718: Got it, that seems reasonable to me

saged/extract_metadata_text.py
ben-heil: Update, it does but I didn't realize because all the instances were removed :facepalm:

saged/utils.py
ben-heil: Yeah that's definitely more clear

saged/predict_tissue.py
ben-heil: Nope, since the individual tissue comparisons (or at least Blood/Breast) seem to be easy, I'm focusing on the multitissue classification

saged/predict_tissue.py
ben-heil: Yes, and I probably should

Also I think this code is wrong since the matrix is genes x samples instead of samples x genes. I should be scaling along the embedding dimension not the sample dimension

PR #52, opened 2021-09-09
Sample split

ben-heil:
I've gotten some encouraging results recently, but my previous project has left me skeptical of the results of any single experiment. To be more confident that transfer learning is actually doing anything, I planned out a series of three experiments that help narrow down exactly what's happening

1) Split the dataset (at the sample level) into pretrain, train, and val splits and run cross-validation. Create two copies of a model with the same initialization. Pretrain one model on the pretrain split, then train both on the training split. Evaluate their performance on the validation split. Since the datasets are split at the sample level (which is to say are roughly iid), if this experiment fails then there is something wrong with my training code or it takes very little data to max out performance.
2) Perform the same experiment, but with the datasets split at the study level so that the train and val sets have the same differences in distribution they would in a real setting. If experiment 1 succeeds but this fails, then inter-study heterogeneity is driving our inability to predict tissue
3) Split the dataset at the study level, but use different tissues in the pretrain and train/val sets. If this experiment fails, then pretraining helps in a tissue-specific manner

This PR implements experiments 1 and 2, with very positive results (see the images at the bottom of `benchmark_results`). Both experiment 1 and 2 are successful, and you can see that performance drops dramatically when you stop leaking information from the val set via sample level splitting

Changes:
- Updated the `datasets.py` code to allow `get_cv_splits` to split at the study level or the sample level
- Wrote `sample_split_control` to run experiments 1 and 2

Questions:
- There seems to be a maximum performance of around 55% balanced accuracy at which point more data doesn't help. I'm unsure of what is causing this, but if you have an idea let me know

### DIFF
--- a/notebook/converted/benchmark_results.py
+++ b/notebook/converted/benchmark_results.py
@@ -4,7 +4,7 @@
 # # Benchmark Results
 # This notebook visualizes the output from the different models on different classification problems
 
-# In[2]:
+# In[1]:
 
 
 import collections
@@ -1295,6 +1295,8 @@ for path in in_files:
     
 tissue_metrics['train_count'] = tissue_metrics['train sample count']
 tissue_metrics = tissue_metrics.rename({'impute_samples': 'pretraining_sample_count'}, axis='columns')
+tissue_metrics['supervised'] = tissue_metrics['supervised'].str.replace('pytorch_supervised', 'three_layer_net')
+
 tissue_metrics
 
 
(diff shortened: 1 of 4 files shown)


saged/datasets.py
```diff
@@ -273,6 +276,12 @@ def get_cv_splits(self, num_splits: int, seed: int = 42) -> Sequence["Expression
         ---------
         num_splits: The number of groups to split the dataset into
         seed: The seed for the random number generator involved in subsetting
+        split_by_sample: If this flag is set, the cross-validation splits will not be study-aware.
+            That is to say that data will leak between the CV splits if the flag
+            is set to True. As a result, it should only be used as a positive control
+            with the understanding of how this fails to be an accurate representation
+            of performance from e.g.
+            autobencoder.com/2021-01-29-paper-evaluation/#failures-of-data-splitting-in-biology
```
jjc2718, line 284: Always fun when you can cite yourself 😂

saged/datasets.py
```diff
@@ -754,31 +772,55 @@ def get_cv_expression(self, num_splits: int, seed: int = 42) -> List[pd.DataFram
...
         samples = self.get_samples()
         studies = self.get_studies()
-        shuffled_studies = utils.deterministic_shuffle_set(studies)
 
-        base_study_count = len(studies) // num_splits
-        leftover = len(studies) % num_splits
-        study_index = 0
+        if split_by_sample:
+            shuffled_samples = utils.deterministic_shuffle_set(samples)
+            base_sample_count = len(samples) // num_splits
```
jjc2718, line 778: I wonder if it's possible to move any of this into a function - after you get the shuffled samples or studies, it looks like most of this logic is the same whether you're splitting by sample or study.

Maybe something to think about in the future, since what you have seems to be working fine for now.

saged/sample_split_control.py
```diff
@@ -0,0 +1,217 @@
...
+
+        model_type = model_config['name']
+        SupervisedClass = getattr(models, model_type)
+        root_model = SupervisedClass(**model_config)
+
+        pretrain_base = copy.deepcopy(root_model)
+
+        # Get the pretraining fraction of the data
+        DatasetClass = type(labeled_data)
+        pretrain_list = [labeled_splits[(i+1) % 5], labeled_splits[(i+2) % 5]]
```
jjc2718, line 109: Do you want to hard-code 5 here, or could you make this an argument to the script?

notebook/converted/benchmark_results.py
```diff
@@ -1313,3 +1315,143 @@ def return_unlabeled():
...
+tissue_metrics['train_count'] = tissue_metrics['train sample count']
+tissue_metrics['supervised'] = tissue_metrics['supervised'].str.replace('pytorch_supervised', 'three_layer_net')
+
+tissue_metrics
+
+
+# In[16]:
+
+
+plot = ggplot(tissue_metrics, aes(x='train_count', y='balanced_accuracy', color='is_pretrained'))
```
jjc2718, line 1357: Why doesn't pretraining work for logistic regression? Or is it just that the baseline performance is so much lower than for the NN that you can't see any real effect?

notebook/converted/benchmark_results.py
```diff
@@ -1313,3 +1315,143 @@ def return_unlabeled():
...
+
+# In[23]:
+
+
+plot = ggplot(single_run_df, aes(x='train_count', y='balanced_accuracy', color='is_pretrained')) 
+plot += geom_smooth()
+plot += geom_point(alpha=.2)
+plot += ggtitle('Study Level Single Run Points')
+plot += facet_grid('supervised ~ .')
+plot
```
jjc2718, line 1450: Could breaking the results down by tissue type help you understand the balanced accuracy 55% maximum? Maybe there are certain tissue types where performance is always poor, and these are effectively putting a cap on how high your accuracy can be overall - just an idea.

jjc2718 (approved):
Looks good to me!

saged/datasets.py
ben-heil: Absolutely, "I explained this better in the past so look here" is one of the best reasons to write a blog

saged/datasets.py
ben-heil: It definitely is. I've been putting off refactoring to see if the code works well enough to be worth keeping. My current plan is to run one more experiment then refactor a number of scripts

saged/sample_split_control.py
ben-heil: I think I'll leave it hardcoded now then maybe update it during refactoring. It's hard to generalize a 2:2:1 ratio to arbitrary input, though to be fair it's only 2:2:1 to parallel the 20% val set other scripts, and they take arbitrary numbers of splits so :shrug:

notebook/converted/benchmark_results.py
ben-heil: That's a good question. I would expect logistic regression to have the same performance in both sets because you're optimizing a concave function with a global optimum. In that sense if training works correctly, the only thing the starting weight change is how long it takes to get there.

On the other hand, my pytorch implementation of logistic regression is helped by pretraining. Maybe the pytorch version isn't given enough epochs to converge, or (my working theory) that the sklearn optimizer is less noisy than batch SGD

PR #53, opened 2021-09-17
Remove scrna

ben-heil:
This is a pretty short PR, it just adds a script that removes scRNA samples from the compendium.

Recount had two methods of labeling scRNAseq. I though their pattern-based one (below) was more believable. However, it seems to call a lot of very sparse RNA-seq in the upper ranges, so I also set a manual 75% sparsity cutoff. 
image


renv.lock
```diff
@@ -0,0 +1,682 @@
+{
```
jjc2718, line 1: Do you like `renv` better than using conda with R packages? I've always found conda impossible to use with R - interested in trying something different next time I have to write some R code.

saged/remove_scrnaseq.py
```diff
@@ -0,0 +1,81 @@
...
+            # Skip malformed lines
+            except TypeError:
+                continue
+
+            try:
+                if sparsity > .75 or recount_pred == 'scrna-seq':
+                    continue
+                else:
+                    out_file.write(line)
+            except ValueError as e:
```
jjc2718, line 79: Why would a `ValueError` happen here? May be worth adding a comment.

saged/remove_scrnaseq.py
```diff
@@ -0,0 +1,81 @@
...
+    # Drop rows without sample ids and duplicate rows
+    metadata = metadata[metadata['external_id'].notna()]
+    metadata = metadata.drop_duplicates(subset=['external_id', 'study'])
+    # Set the index to the sample id for faster access
+    metadata = metadata.set_index('external_id')
+
+    LINES_IN_FILE = 317259
+
+    with open(args.count_file, 'r') as count_file:
+        out_file = open(args.out_file, 'w')
```
jjc2718, line 29: Do you want to close this file handle somewhere?

saged/remove_scrnaseq.py
```diff
@@ -0,0 +1,81 @@
...
+            except ValueError:
+                continue
+
+            # This works for int zeros and float zeros
+            zero_count = counts.count(0)
+            sparsity = zero_count / total_genes
+            try:
+                sample_metadata = metadata.loc[sample, :]
+            except KeyError:
+                # print(e)
```
jjc2718: Can you get rid of this line? I'm guessing it was for debugging.

jjc2718 (approved):

renv.lock
ben-heil: Yeah, it's great! (Or at least it's great so far, I haven't tried to load an renv environment on another machine so we'll see). On the download side it mostly just works like I want Conda to e.g. it automatically detects the difference between bioconductor and CRAN packages

saged/remove_scrnaseq.py
ben-heil: ValueErrors crop up when a line is malformed but still passes through `len`. Went ahead and added a comment

saged/remove_scrnaseq.py
ben-heil: Good call

saged/remove_scrnaseq.py
ben-heil: Right, good catch!

PR #55, opened 2021-09-27
Tissue split

ben-heil:
This PR adds a script to determine whether pretraining expression models on different tissues increases performance, and the answer seems to be no.

I'm somewhat skeptical of the results, since the performance doesn't increase as well as I'd expect when data increases. We'll see whether this holds up after I rerun everything with the scRNA removed, but I'd expect it to. If true, this is an interesting result, it says "transfer learning works, but there are limits"

image

Changes
- Add tissue_split.py to run the analysis
- Update the results notebook
- Update the Snakefile
- Add a line to the model config files to allow tissue_split to reinitialize their last layer's parameters when running tissue_split

### DIFF
--- a/notebook/converted/benchmark_results.py
+++ b/notebook/converted/benchmark_results.py
@@ -1450,8 +1450,66 @@ plot += facet_grid('supervised ~ .')
 plot
 
 
-# In[ ]:
+# ## Tissue Split
+
+# In[2]:
 
 
+in_files = glob.glob('../../results/tissue-split.*.tsv')
+print(in_files[:5])
 
+
+# In[3]:
+
+
+tissue_metrics = pd.DataFrame()
+
+for path in in_files:
+    new_df = pd.read_csv(path, sep='\t')
+    model_info = path.strip('.tsv').split('biobert.')[-1]
+    model_info = model_info.split('.')
+    model_info = model_info[-1]
+    model_info = model_info.split('_')
+        
+    supervised_model = '_'.join(model_info[:2])
+             
+    new_df['supervised'] = supervised_model
+    
+    new_df['seed'] = model_info[-1]
+            
+    tissue_metrics = pd.concat([tissue_metrics, new_df])
+    
+tissue_metrics['train_count'] = tissue_metrics['train sample count']
+tissue_metrics['supervised'] = tissue_metrics['supervised'].str.replace('pytorch_supervised', 'three_layer_net')
+
+tissue_metrics
+
+
+# In[4]:
+
+
+plot = ggplot(tissue_metrics, aes(x='train_count', y='balanced_accuracy', color='is_pretrained')) 
+plot += geom_smooth()
+plot += geom_point(alpha=.2)
+plot += ggtitle('Cross-Tissue Pretraining')
+plot += facet_grid('supervised ~ .')
+plot
+
+
+# In[5]:
+
+
+single_run_df = tissue_metrics[tissue_metrics['seed'] == '1']
+single_run_df.head()
+
+
+# In[6]:
+
+
+plot = ggplot(single_run_df, aes(x='train_count', y='balanced_accuracy', color='is_pretrained')) 
+plot += geom_smooth()
+plot += geom_point(alpha=.2)
+plot += ggtitle('Cross-tissue Single Run Points')
+plot += facet_grid('supervised ~ .')
+plot
 
--- a/saged/tissue_split.py
+++ b/saged/tissue_split.py
@@ -1,0 +1,222 @@
+"""
+This experiment determines whether models pretrained on data from different tissues than they
+are predicting do better than models without pretraining
+"""
+import argparse
+import copy
+import gc
+import random
+
+import sklearn.metrics
+import torch
+
+import yaml
+
+from saged import utils, datasets, models
+from saged.models import LogisticRegression
+
+
+PREDICT_TISSUES = ['Blood', 'Breast', 'Stem Cell', 'Cervix', 'Brain', 'Kidney',
+                   'Umbilical Cord', 'Lung', 'Epithelium', 'Prostate', 'Liver',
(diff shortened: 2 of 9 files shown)


saged/tissue_split.py
```diff
@@ -0,0 +1,230 @@
+"""
+This experiment determines whether models pretrained on data from different tissues than they
```
jjc2718, line 2: Just so I understand - before (e.g. in #52), you were pretraining on data from a mixture of tissues including the one you were trying to predict expression in? And now you're pretraining on one tissue (or a mixture of tissues not including the holdout tissue), and predicting on a different one?

saged/tissue_split.py
```diff
@@ -0,0 +1,230 @@
...
+import sklearn.metrics
+import torch
+
+import yaml
+
+from saged import utils, datasets, models
+from saged.models import LogisticRegression, PytorchSupervised
+
+
+PREDICT_TISSUES = ['Blood', 'Breast', 'Stem Cell', 'Cervix', 'Brain', 'Kidney',
```
jjc2718, line 19: Seems like these `PREDICT_TISSUES` are the same across all of your tissue pretraining experiments - maybe you could put this list somewhere centralized and import/load it in each script? Unless you want to set it individually in each script.

saged/utils.py
```diff
@@ -54,6 +55,93 @@
...
+        data = data.subset_samples_to_labels(subset)
+        # This is very memory intensive since the full dataset gets stored behind the scenes after
+        # subsetting. It could be made more efficient with a special subset function that
+        # removes the old data, but that may make the API too messy
+        data_copy = copy.deepcopy(data)
+        data.reset_filters()
+
+        splits.append(data_copy)
+
+    # Ensure no studies are shared between the datasets
```
jjc2718, line 120: Are there a lot of studies that contain samples from multiple tissues?

saged/tissue_split.py
```diff
@@ -0,0 +1,230 @@
...
+
+                print('Samples: {}'.format(len(train_data.get_samples())))
+                print('Studies: {}'.format(len(train_data.get_studies())))
+
+                output_size = len(train_data.get_classes())
+
+                print('Val data: {}'.format(len(val_data)))
+                print('output size: {}'.format(output_size))
+
+                # TODO modify save path for no_pretraining_model
```
jjc2718, line 170: Do you want to keep this TODO here?

notebook/converted/benchmark_results.py
```diff
@@ -1450,8 +1450,66 @@ def return_unlabeled():
...
+tissue_metrics['train_count'] = tissue_metrics['train sample count']
+tissue_metrics['supervised'] = tissue_metrics['supervised'].str.replace('pytorch_supervised', 'three_layer_net')
 
+tissue_metrics
+
+
+# In[4]:
+
+
+plot = ggplot(tissue_metrics, aes(x='train_count', y='balanced_accuracy', color='is_pretrained'))
```
jjc2718, line 1491: I'm surprised that there's not more heterogeneity/variance between splits. I'd assume that transfer between some (related) tissues would work well, but other tissues would be too different to share much information. But it looks here like none of your train/test splits really work well (which I guess is plausible, but surprising at least to me).

notebook/converted/benchmark_results.py
```diff
@@ -1450,8 +1450,66 @@ def return_unlabeled():
...
+tissue_metrics['supervised'] = tissue_metrics['supervised'].str.replace('pytorch_supervised', 'three_layer_net')
 
+tissue_metrics
+
+
+# In[4]:
+
+
+plot = ggplot(tissue_metrics, aes(x='train_count', y='balanced_accuracy', color='is_pretrained')) 
+plot += geom_smooth()
```
jjc2718, line 1492: Do you have any idea why logistic regression seems to do better here than the NN models?

jjc2718 (approved):
Code looks good! Interesting results, unfortunate that cross-tissue transfer doesn't seem to work too well but I guess it's not that surprising that there are limits.

saged/tissue_split.py
ben-heil: Correct, the pretraining data is a different set of studies with different samples than the training and validation sets. The train/val sets have the same tissues, but are still split by study

saged/tissue_split.py
ben-heil: Yeah, that's a good point, I'll save that change for the refactoring PR though

saged/utils.py
ben-heil: There are a decent amount, iirc I lose ~10 percent of the data by making sure that the studies are distinct across splits

saged/tissue_split.py
ben-heil: Yeah, it isn't critical currently, but I'll want to have example models available if/when this goes into a paper

notebook/converted/benchmark_results.py
ben-heil: There are actually only two tissue splits (mutiplying model_types * pretrain_or_random * cv_splits * initializations * data subsets by another factor would make the experiment take even longer). I'd expect that there are splits that would perform better, that may be an interesting future experiment (e.g. pretraining based on tissue similarity).

notebook/converted/benchmark_results.py
ben-heil: I have no idea. I think there may be a bug in the implementation causing the accuracies to be uniformly low (I'd expect a non-pretrained model to have ~40% accuracy). As for why LR does better, I don't know

notebook/converted/benchmark_results.py
jjc2718: Makes sense!

notebook/converted/benchmark_results.py
jjc2718: Yeah, the NN accuracies being so low across the board is a bit surprising. Something to investigate in the future, maybe.

PR #56, opened 2021-09-29
No scrna

ben-heil:
This PR is an easy one to review since there aren't any code changes :smile:. It visualizes results from experiments run on the recount dataset after the single cell RNA-seq data had been removed.

The results are mostly the same (as far as I've noticed), but with generally increased accuracy. Questions remain:

1) Why is logistic regression so good in the sample split (i.e. data leaking) case?
2) Why does LR outperform in the cross-tissue case?
3) Do these results mean that DL just isn't useful in gene expression classification tasks?

Old sample split:
image

New sample split:
image

Old study split:
image

New study split:
image

Old tissue split:
image

New tissue split:
image

### DIFF
--- a/notebook/converted/benchmark_results.py
+++ b/notebook/converted/benchmark_results.py
@@ -1318,14 +1318,14 @@ plot
 
 # ## Sample Split Positive Control
 
-# In[14]:
+# In[2]:
 
 
 in_files = glob.glob('../../results/sample-split.*.tsv')
 print(in_files[:5])
 
 
-# In[15]:
+# In[3]:
 
 
 tissue_metrics = pd.DataFrame()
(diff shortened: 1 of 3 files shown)


notebook/converted/benchmark_results.py
```diff
@@ -1485,7 +1485,7 @@ def return_unlabeled():
 tissue_metrics
 
 
-# In[4]:
+# In[9]:
```
jjc2718, line 1488: Do you want to plot the results with and without scRNA-seq samples removed here, instead of overwriting the old ones? Or are you sure you'll only use the dataset with scRNA-seq samples removed going forward?

jjc2718 (approved):
Looks good! Makes sense that accuracy gets better across the board when single cell samples are removed.

> 3. Do these results mean that DL just isn't useful in gene expression classification tasks?

I hope this isn't true! Although I guess it's possible that in many cases the data is noisy/collinear enough that a linear-ish model makes the most sense.

notebook/converted/benchmark_results.py
ben-heil: I plan on only using the dataset without scRNA-seq samples going forward.

PR #59, opened 2021-11-04
Error analysis

ben-heil:
This PR runs some analyses to try to figure out what is causing the irreducible error in model performance so far (hint, it's always label imbalance)

Changes:
- Add logic to sample split and tissue split experiments to record what predictions are made by each iteration of each model
- Add an analysis notebook digging into which samples are hard to classify and why


jjc2718 (approved):
Looks good! This is a nice, thorough analysis.

Comments on `notebook/analysis/visualize_errors.ipynb` (can't add them inline since the file is too large):

* It seems like there's a lot of logic repeated between cell 10 and cell 80 (i.e. for building the confusion matrices). Would it be possible to put some of this into a function in a utilities file, or is the goal of these analyses considerably different?

* In the confusion matrix in cell 14, is it correct that most placenta samples are being classified as lung tissue, and only 2 are classified correctly? I guess maybe that's an example of the label imbalance you're talking about (there are just so many lung samples in the dataset that your model tends to predict that very frequently).

* Re. the sequencer results (cell 73 and on): did you look at sequencer proportion by tissue? Maybe it's just that the difficult tissues tended to use the Hi-seq 2000 and other sequencers that are enriched for hard samples. I guess tissue type probably tracks pretty closely with study, though, so adding tissue type to the analysis might not add much.

* Have you tried just plotting tissue proportion in the dataset vs. model accuracy on that tissue? Given these results my guess is that you should see a positive correlation (higher frequency -> better accuracy), but presumably you want to see less correlation once you tweak things (i.e. if your dataset is big enough and your model is picking up on infrequent labels correctly, there shouldn't be as much correlation between frequency of tissue type and accuracy). Could be a simple way to evaluate whether whatever you change in the future is having the effect you want.

ben-heil:
> In the confusion matrix in cell 14, is it correct that most placenta samples are being classified as lung tissue, and only 2 are classified correctly? I guess maybe that's an example of the label imbalance you're talking about (there are just so many lung samples in the dataset that your model tends to predict that very frequently).

Yep, that is correct! There are very few placental samples and they are consistently misclassified

I'm putting together another PR tomorrow that modifies this notebook to include the updated data, so I'll add the suggested improvements there

PR #60, opened 2021-11-12
Loss weighting

ben-heil:
I weighted the loss functions by inverse class frequency, which improved the overall accuracy, but didn't change anything with respect to the error distribution being bimodal

Changes:
- Updated `benchmark_results` and `visualize_errors` to have the results of the new runs
- Updated the `PytorchSupervised` in `models.py` to use a weighted loss (and in doing so resolved a 16 month old TODO)
- Updated tissue split and sample split to weight losses
- Added a function to `utils` to calculate the loss weights


notebook/converted/visualize_errors.py
```diff
@@ -0,0 +1,848 @@
...
+pca_df['is_hard'] = is_hard
+
+
+# In[131]:
+
+
+plot = ggplot(pca_df, aes(x='UMAP1', y='UMAP2', fill='is_hard'))
+plot += geom_point()
+plot += ggtitle('UMAP Embedding of all samples')
+print(plot)
```
jjc2718: This is actually plotting the PCA embedding, right? Maybe you copied over your previous plotting code and forgot to change the embedding name.

notebook/converted/visualize_errors.py
```diff
@@ -0,0 +1,848 @@
...
+
+melted_df = confusion_df.melt(id_vars='true_tissue', var_name='pred_tissue')
+melted_df['log_value'] = np.log(melted_df['value'])
+melted_df['percent'] = (melted_df['value'] * 100).round(1)
+
+
+# In[164]:
+
+
+plot = ggplot(melted_df, aes(x='true_tissue', y='pred_tissue', fill='percent',))
```
jjc2718, line 835: Have you thought about whether this confusion matrix makes biological sense? For example, maybe it makes sense that some blood cells are transcriptionally closer to stem cells, or that some brain cells look like skin cells (I don't know enough about development/cell lineages to say whether these are true, just ideas that I see in your plot).

jjc2718 (commented):
Code looks good to me!

Any idea why the tests are failing?

notebook/converted/visualize_errors.py
ben-heil: Good catch! Fixed it and some other umap -> pca swaps

notebook/converted/visualize_errors.py
ben-heil: I would expect that to be part of the signal, but the outliers seem unrelated. For example, everything gets predicted to be lung tissue or spinal cord.  While I'm not sure what causes those horizontal streaks, some of the errors I think are biological, like the bone marrow <-> blood mixups

ben-heil:
> Any idea why the tests are failing?

Because I didn't make any changes to the API so I didn't remember to check the tests :)

The tl;dr is that I forgot some quotes to do a forward reference on one of the types in `calculate_loss_weights`.
The version that is way longer and ate one of my afternoons about a year ago is:

The error is because I referenced the `saged.datasets.RefineBioLabeledDataset` class as a normal type and `utils.py` doesn't know what that type is. 

"But Ben, you import saged.datasets at the top of `utils`, how does that make any sense?"

Well `datasets.py` is imported first, which imports `utils.py` in its header. Since python is an interpreted language, the function `calculate_loss_weights` is read before the `RefineBioLabeledDataset`.

"But Ben,  doesn't `utils.py` import `datasets` in its header?"

Yes, and to avoid the infinite loop that that entails, python automatically avoids importing files which have already been imported.

"But Ben, isn't it bad code architecture to have two files that refer to each other? Shouldn't `datasets` reference `utils` but not the other way around?"

Yes, I think so, but the only reason `datasets` is imported in `utils` is to allow dataset types in function signatures. Under that construction there would always be a circular reference, so the type resolution would always fail unless I manually ensure that `utils` is always imported before `datasets` in every script.

"That sounds like a bad hack, but shouldn't there be a way to just reference types that will be declared later ?"

Yep, and that's what [forward references](https://www.python.org/dev/peps/pep-0484/#forward-references) are for, and why I forgetting some quotation marks caused the tests to fail.

jjc2718:
Sounds good! Nice exploration and explanation of the issue.

PR #61, opened 2021-11-21
Standardize

ben-heil:
As I start wrapping up Saged, I wanted to make sure that all the results are comparable so that it's clear what is happening. To make that a reality, I reran all the tissue prediction steps with rna-seq only data and loss weighting.

Changes:
- Removed old parts of README (the current version definitely isn't the final one, but it's at least closer)
- Updated the Snakefile to use weighted loss
- Removed old transformer configs
- Added a save path for the sklearn logistic regression model
- Reran the jupyter notebook for visualizing results
- Added SVA and Limma to renv 
- Updated imputation_pretraining and predict_tissue to use weighted losses
- Tweaked models.py to make early stopping more patient
- Updated sample_split control to save models


README.md
```diff
@@ -1,31 +1,14 @@
 # SAGED
 
 ![Status Badge](https://github.com/greenelab/saged/workflows/PythonTests/badge.svg)
-[![Coverage Status](https://coveralls.io/repos/github/ben-heil/saged/badge.svg?branch=master)](https://coveralls.io/github/ben-heil/saged?branch=master)
```
jjc2718, line 4: Do you want some kind of description in the README of what you're doing? Or would you add that later?

saged/models.py
```diff
@@ -677,7 +677,7 @@ def __init__(self,
                  save_path: str = None,
                  train_fraction: float = None,
                  train_count: float = None,
-                 early_stopping_patience: int = 3,
+                 early_stopping_patience: int = 7,
```
jjc2718, line 680: Does changing this affect your results?

jjc2718 (approved):
Looks good!

README.md
ben-heil: I definitely will, at some point in the future I'm going to do a big housekeeping PR where I rewrite the README and remove unused code/files

saged/models.py
ben-heil: Not really, the difference was e.g. from the first image to the second image. I mostly just did it to try to make sure models won't quit due to instability in the beginning

image
image

PR #62, opened 2021-11-28
Sex Prediction

ben-heil:
To ensure the results that we found aren't just artifacts of trying to predict tissue, we decided to predict the sex labels extracted from the metadata by [Flynn et al.](https://bmcbioinformatics.biomedcentral.com/articles/10.1186/s12859-021-04070-2)

We found the same patterns to be present in sex prediction, even though it's an easier task overall.

Changes:
- Add sex prediction rules to Snakefile
- Visualize results in benchmark_results.ipynb
- Visualize the compendium data in lower dim space (recount_umap.ipynb)
- Add an argument to sample_split_control.py to allow sex prediction instead of tissue prediction
- Added a function to utils.py to parse the Flynn et al. labels


notebook/analysis/benchmark_results.ipynb
```diff
@@ -10517,12 +10517,1360 @@
     "plot"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Sex Prediction"
+   ]
+  },
```
jjc2718, line 10526: Do you want something in this cell? I just see an empty cell after the "Sex Prediction" heading.

notebook/analysis/benchmark_results.ipynb
```diff
@@ -10517,12 +10517,1360 @@
...
    "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": []
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Sample Split Positive Control"
```
jjc2718, line 10538: Are these results fairly comparable to what they found in the paper you linked, when they did prediction from gene expression? Or are the experiments you're doing substantially different so you can't compare?

saged/sample_split_control.py
```diff
@@ -49,17 +49,22 @@
     parser.add_argument('--weighted_loss',
                         help='Weight classes based on the inverse of their prevalence',
                         action='store_true')
+    parser.add_argument('--sex_label_path',
```
jjc2718: So if this argument is included, the label/target variable is sex, and if it's not the target variable is tissue type, correct? Might be worth mentioning that this argument controls whether sex labels are used, and not just where they are located.

jjc2718 (approved):
Do you want to add a download script for the sex labels, or some directions in the README on where to find them? If you're going to (eventually) put this result in a paper it would be good to have some direction on where the data lives, but if you're not sure maybe it's fine to wait.

notebook/analysis/benchmark_results.ipynb
ben-heil: Yeah, both of our accuracies top out around 90%. (Which is interesting in and of itself because they're using only X and Y chromosome genes)

saged/sample_split_control.py
ben-heil: That's right, and I agree that that's a weird design choice. I'll change it to a boolean flag, then retrieve the path to the labels from the dataset config file

notebook/analysis/benchmark_results.ipynb
ben-heil: Nope, added a commit to remove the empty cell

ben-heil:
> Do you want to add a download script for the sex labels, or some directions in the README on where to find them? If you're going to (eventually) put this result in a paper it would be good to have some direction on where the data lives, but if you're not sure maybe it's fine to wait.

I actually screwed up and [added the instructions to the previous PR](https://github.com/greenelab/saged/pull/61/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R5). I would add a download script, but the labels are in a private figshare repo so I haven't found a permalink that I could put in a script

PR #63, opened 2021-12-01
Label correction

ben-heil:
We decided to look at the results when the linear tissue signal was removed from samples via Limma before prediction. In theory, deep networks should be able to learn nonlinear signal and do better, but the 21-class tissue prediction problem may be too difficult.

Changes:
- Added relevant rules to Snakefile
- Removed old disease-prediction logic from benchmark_Results
- Visualized results in benchmark_results
- Added logic to remove study and tissue signals to sample_split_control

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -80,6 +80,16 @@ rule all:
                supervised=SUPERVISED,
                seed=range(0,NUM_SEEDS),
                ),
+        # sample_split be_corrected
+        expand("results/sample-split-study-corrected.{supervised}_{seed}.tsv",
+               supervised=SUPERVISED,
+               seed=range(0,NUM_SEEDS),
+               ),
+        # study_split be_corrected
+        expand("results/study-split-study-corrected.{supervised}_{seed}.tsv",
+               supervised=SUPERVISED,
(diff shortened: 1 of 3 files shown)


jjc2718 (approved):
Looks good to me!

Interesting that for the sample level experiments, the `pytorch_lr` and `deep_net` models do seem to benefit slightly from increased samples - do you think that's just an artifact of higher `train_count` values being undersampled, or the balanced accuracy baseline increasing with more samples? Or can you interpret this as suggesting there might be some non-linear signal in the data, but it just requires lots of samples to identify it?

ben-heil:
> 



> Looks good to me!
> 
> Interesting that for the sample level experiments, the `pytorch_lr` and `deep_net` models do seem to benefit slightly from increased samples - do you think that's just an artifact of higher `train_count` values being undersampled, or the balanced accuracy baseline increasing with more samples? Or can you interpret this as suggesting there might be some non-linear signal in the data, but it just requires lots of samples to identify it?

My intuition is more towards the latter. I think the binary classification results (will PR a lot more of those later today) seem to bear out that there is some nonlinear signal, it's just hard to tell in a 21 class problem with only a few thousand samples

PR #64, opened 2021-12-04
Binary classification

ben-heil:
This PR does a couple of things (sorry). The primary one is to run an experiment to determine whether deep models outperform linear models when the linear signal has been removed (they do). It also adds inverse class weighting to the scikit-learn version of LR and adds a notebook to try to figure out why LR does worse than random when signal is removed.

Changes:
- Updated snakefile to do pairwise binary classification instead of doing a specific tissue pair
- Create notebook (bad_lr_performance) to plot confusion matrices etc of the signal removed models
- Add plots to benchmark_results for pairwise comparisons (if you can think of a better way of showing things than a 2d facet grid, let me know
- Tweak a few things in predict_tissue to make pairwise comparisons work, enable both types of BE correction, and make two-word tissues play nicely with argparse


notebook/converted/bad_lr_performance.py
```diff
@@ -0,0 +1,382 @@
...
+from plotnine import *
+
+from saged.utils import split_sample_names, create_dataset_stat_df, get_dataset_stats, parse_map_file
+
+
+# In[3]:
+
+
+top_five_tissues = ['Blood', 'Breast', 'Stem_Cell', 'Cervix', 'Brain']
+#top_five_tissues = ['Blood', 'Breast', 'Stem_Cell']
```
jjc2718, line 28: Can you get rid of this line?

notebook/converted/bad_lr_performance.py
```diff
@@ -0,0 +1,382 @@
...
+        new_df['tissue2'] = tissue2
+    else:
+        new_df['tissue1'] = tissue2
+        new_df['tissue2'] = tissue1
+    new_df['correction_method'] = correction_method
+    new_df['pair'] = new_df['tissue1'] + '-' + new_df['tissue2'] 
+    
+    run_results = pd.concat([run_results, new_df])
+    
+run_results['train_count'] = run_results['train sample count']
```
jjc2718, line 78: Can you use [the rename function](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.rename.html) for this? That would also get rid of the original column, which it seems like is what you want here.

notebook/converted/bad_lr_performance.py
```diff
@@ -0,0 +1,382 @@
...
+    predicted = model_predictions[model]['predicted']
+    true = model_predictions[model]['true']
+
+    confusion_matrix = sklearn.metrics.confusion_matrix(true, predicted)
+    
+    tissue_set = set(true).union(set(predicted))
+    tissues = sorted(list(tissue_set))
+    
+    if norm:
+        confusion_matrix = np.apply_along_axis(row_norm, axis=1, arr=confusion_matrix)
```
jjc2718, line 164: Can you use the [`normalize` argument of `sklearn.metrics.confusion_matrix`](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.confusion_matrix.html) to do this?

notebook/converted/bad_lr_performance.py
```diff
@@ -0,0 +1,382 @@
...
+    confusion_df = pd.DataFrame(confusion_matrix, index = [l for l in tissues],
+                                columns = [l for l in tissues])
```
jjc2718, line 166: Do you need the list comprehensions here? It looks like you already convert `tissues` to a list above, so it seems like you should just be able to pass it directly.

notebook/converted/bad_lr_performance.py
```diff
@@ -0,0 +1,382 @@
...
+plot = ggplot(all_df, aes(x='factor(pred_or_true)', y='Brain', fill='model'))
+plot += geom_bar(stat='identity')
+plot += facet_grid('model ~ .')
+print(plot)
+
+
+# ## Conclusion
+# There doesn't seem to be anything significantly different between the prediction distributions of the deep models and the linear models. The linear ones are just wrong more regardless of class.
+# 
+# Actually the differences between the five layer net and three layer net in terms of their prediction strategy are interesting (they perform better with respect to different classes).
```
jjc2718, line 380: Maybe if you used them to build an ensemble you could do even better! 😂 

(I know maximizing performance isn't really your goal here, mostly just joking - but it is interesting that they perform so differently)

jjc2718 (approved):
Looks good to me! Just a few small comments.

Comments on `benchmark_results.ipynb`:

* I think the facet grid looks fine, but is there any way you can make the individual plots (facets) have different x-axes? It's hard to see the difference between models for some of the pairs because the maximum sample size is so small.
* Are you surprised at how well the nonlinear models do when the tissue signal is removed? I guess it's only really a strong effect in the blood -> breast case, the nonlinear models don't do as well in the other cases. I'm seeing a similar effect in some of my experiments (PR to come soon) - just curious if that's what you expected, or if you're surprised by that.

notebook/converted/bad_lr_performance.py
ben-heil: Yep!

notebook/converted/bad_lr_performance.py
ben-heil: Good catch!

notebook/converted/bad_lr_performance.py
ben-heil: It's possible, but iirc there is something confusing about whether the rows or columns are predictions, so I'm afraid to mess with it and screw it up

notebook/converted/bad_lr_performance.py
ben-heil: Yeah probably! I'm surprised too, I'd expect the same model class to have the same strategy for optimizing the loss, but apparently not

ben-heil:
> Looks good to me! Just a few small comments.
> 
> Comments on `benchmark_results.ipynb`:
> 
> * I think the facet grid looks fine, but is there any way you can make the individual plots (facets) have different x-axes? It's hard to see the difference between models for some of the pairs because the maximum sample size is so small.

I didn't think it was, but it turns out it is. I'm currently rerunning all the binary classification due to unnecessary subsetting, but I'll update the grids in the next PR

> * Are you surprised at how well the nonlinear models do when the tissue signal is removed? I guess it's only really a strong effect in the blood -> breast case, the nonlinear models don't do as well in the other cases. I'm seeing a similar effect in some of my experiments (PR to come soon) - just curious if that's what you expected, or if you're surprised by that.

I guess it's not really surprising to me. My intuition would be that there is nonlinear signal relating tissues to gene expression, so it makes sense that DL models can learn that signal. The surprising part (to me at least), is that the nonlinear signal seems much less important than the linear signal in predicting tissue

PR #65, opened 2021-12-22
Optimize Hyperparameters

ben-heil:
This PR implements nested cross-validation for setting hyperparameters. Doing so takes ~10x longer to run, so only the all_tissue_prediction script has been rerun for now.

Results:
Things look pretty much the same; hyperparameter optimization doesn't matter too much. First image is manual hyperparameters, second is optuna

image
image

Changes:
- Update model configs to have a 'l2_penalty' field to optimize via optuna
- Visualize results (all tissue prediction figure)
- Add a `predict_proba` function to the model classes to give a way to return probabilities not just predictions
- Update `predict_tissue` to implement hyperparameter optimization (`objective` function probably needs a better name, and will be moved to utils in the future


saged/models.py
```diff
@@ -206,9 +222,12 @@ def __init__(self,
         Arguments
         ---------
         seed: The random seed to use in training
+        lr: The learning rate to be used in training
```
jjc2718: Shouldn't this be something related to the `l2_penalty`?

Also it might be good to mention that this is actually the _inverse_ of the standard L2 penalty (see the documentation for `C`[here](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html)). This shouldn't affect your results much if you're using an automatic optimization framework like optuna, just good to keep in mind.

saged/predict_tissue.py
```diff
@@ -28,6 +30,70 @@
...
+
+def objective(trial, train_list, supervised_config,
+              label_encoder, weighted_loss=False, device='cpu'):
+    losses = []
+
+    with open(supervised_config) as supervised_file:
+        supervised_config = yaml.safe_load(supervised_file)
+    supervised_model_type = supervised_config.pop('name')
+
+    # TODO there should be a better way to do this which allows future skl models
```
jjc2718, line 42: But some sklearn models do have a learning rate parameter, right? I think hard-coding this is fine, unless there's a way to check if the model has a learning rate parameter directly (I can't personally think of a good way, so I'd probably just do what you're doing).

saged/predict_tissue.py
```diff
@@ -28,6 +30,70 @@
...
+        input_size = len(inner_train_data.get_features())
+        output_size = len(label_encoder.classes_)
+
+        with open(args.supervised_config) as supervised_file:
+            supervised_config = yaml.safe_load(supervised_file)
+            supervised_config['input_size'] = input_size
+            supervised_config['output_size'] = output_size
+            supervised_config['log_progress'] = False
+            supervised_config['l2_penalty'] = l2_penalty
+            if supervised_model_type != 'LogisticRegression':
```
jjc2718, line 70: Maybe you could define a list of models that don't have the `lr` parameter in one place (like at the top of the function), then just check against it both here and on line 42?

Just an idea - you only have this check in two places so it's not a huge deal, I'm fine with keeping it like this if you want.

saged/predict_tissue.py
```diff
@@ -156,7 +243,28 @@
...
+
     for i in range(len(labeled_splits)):
+        # Select hyperparameters via nested CV
+        train_list = labeled_splits[:i] + labeled_splits[i+1:]
+
+        if not args.disable_optuna:
+            # optuna.logging.set_verbosity(optuna.logging.ERROR)
+            sampler = optuna.samplers.RandomSampler(seed=args.seed)
+            study = optuna.create_study()
+            print('Tuning hyperparameters...')
```
jjc2718, line 257: How much longer does this typically take (compared to just running your experiments with fixed hyperparameters)?

jjc2718 (approved):
Looks good!

Have you compared the parameters that optuna finds with the default ones that you've been using? I'm curious if you're getting similar performance before/after optimization because your default parameters were pretty close to the best ones, or if it just doesn't matter much what parameters you use for this problem.

saged/models.py
ben-heil: Good catch, I deleted the wrong comment

saged/predict_tissue.py
ben-heil: > But some sklearn models do have a learning rate parameter, right?

True

> I think hard-coding this is fine, unless there's a way to check if the model has a learning rate parameter directly

Yeah, it's a hard fix to implement. I'll fix it if it becomes a problem but not until then

saged/predict_tissue.py
ben-heil: Good idea, I'll add it if I need to include more than one model type. As it is right now it's very clearly bad, where if I was comparing to a const I might think that I had a good solution :)

saged/predict_tissue.py
ben-heil: Roughly 10x as long. Running this took a week where it would normally take an afternoon or night

ben-heil:
> Have you compared the parameters that optuna finds with the default ones that you've been using? I'm curious if you're getting similar performance before/after optimization because your default parameters were pretty close to the best ones, or if it just doesn't matter much what parameters you use for this problem.

IIrc they're within an order of magnitude or two (except the l2 norm, which defaults to zero when not doing hyperparameter opt but ends up small but nonzero after optuna)

PR #67, opened 2022-01-06
Gtex

ben-heil:
This commit has two main parts
1) Adds GTEx data processing/running code to see if the results hold on a new dataset 
2) Swaps sklearn optimizers for LR, and tries full data training for pytorch LR

Results:
Very pretty graph with good inter-model separation from GTEx
image

Different optimizer/training methods on LR don't make much of a difference; the skl vs pytorch gap remains
image

Changes:
- Adds GTEx logic to snakefile
- Creates a config for the GTEx dataset
- Changes the LogisticRegression class to allow different optimizers to be set in the config file
- Adds `download_gtex_data` to do what it says on the box
- Implements an option to let pytorch logistic regression to use full data updates instead of minibatch gradient descent
- Creates data preprocessing script `normalize_gtex.py`
- Refactors `predict_tissue` to separate out data loading logic for recount and gtex
- Adds util functions for parsing gtex datafiles and doing class weighting for skl models


saged/normalize_gtex.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def subset_data(expression_df: pd.DataFrame, n_genes) -> pd.DataFrame:
+    """
+    Subset data to the n most variable genes in the dataset
+
+    Arguments
+    ---------
+    expression_df: A samples x genes dataframe containing the expression data
+
```
jjc2718:
```suggestion
    expression_df: A samples x genes dataframe containing the expression data
    n_genes: The number of genes to subset to
```

saged/predict_tissue.py
```diff
@@ -323,6 +365,10 @@ def objective(trial, train_list, supervised_config,
                 if args.weighted_loss:
                     loss_weights = utils.calculate_loss_weights(train_data)
                     supervised_config['loss_weights'] = loss_weights
+                    if (supervised_config['name'] == 'BatchLogRegSKL' or
+                       supervised_config['name'] == 'LogRegSKL'):
```
jjc2718: Do these exist anymore? It looks like you got rid of them.

saged/utils.py
```diff
@@ -501,6 +542,32 @@ def calculate_loss_weights(dataset: "datasets.RefineBioLabeledDataset") -> torch
     return weights
 
 
+def calculate_skl_class_weights(dataset: "datasets.RefineBioLabeledDataset") -> Dict[int, float]:
```
jjc2718, line 545: Just for future reference, I think you might be doing the same thing here as the `class_weight='balanced'` option that's built into sklearn. Probably would be good to check that it's the same if you decide to use that option, though (I may be totally wrong).

jjc2718 (approved):
Looks good! Still a bit surprised that sklearn logistic regression does so much better than your pytorch implementation, but at least it's consistent across datasets.

saged/normalize_gtex.py
ben-heil: Good catch!

saged/predict_tissue.py
ben-heil: I guess not, the new versions used the pytorch and skl classes as their parents. Will remove this section

saged/utils.py
ben-heil: That's right! The issue with `class_weight='balanced'` is that it doesn't work with the `partial_fit` function. As a result, I wrote this function to implement the same weighting scheme for models using `partial_fit` (I don't have any at the moment, but the first versions of `sag_skl` and `sgd_lr` did)

PR #68, opened 2022-01-11
Gtex binary

ben-heil:
This PR implements pairwise tissue predictions on GTEX data. As it turns out, when your data is clean and consistent, it's easy to tell the difference between brains, skin, esophagus, etc.

image

Changes:
- Add relevant rule to snakefile (and disambiguate wildcards from the recount binary prediction rule)
- Add the binary tissue subsetting logic to the `prep_gtex_data` function in `predict_tissues`
- Visualize the results


Snakefile
```diff
@@ -9,11 +9,13 @@ SEMISUPERVISED, = glob_wildcards("model_configs/semi-supervised/{semisupervised}
...
 
-top_five_tissues = ['Blood', 'Breast', 'Stem_Cell', 'Cervix', 'Brain']
-
-combo_iterator = itertools.combinations(top_five_tissues, 2)
-TISSUE_STRING = ['.'.join(pair) for pair in combo_iterator]
+recount_top_five_tissues = ['Blood', 'Breast', 'Stem_Cell', 'Cervix', 'Brain']
+combo_iterator = itertools.combinations(recount_top_five_tissues, 2)
+RECOUNT_TISSUE_STRING = ['.'.join(pair) for pair in combo_iterator]
 
+gtex_top_five_tissues = ['Blood', 'Brain', 'Skin', 'Esophagus', 'Blood_Vessel']
```
jjc2718, line 16: Are these the top five in terms of sample count?

jjc2718 (approved):
Looks good! Just one small question.

Snakefile
ben-heil: Yep! There isn't really a study equivalent in GTEx (that I know of) so I went with sample count

PR #69, opened 2022-01-14
Data simulation solving the negative accuracy mystery

ben-heil:
In previous PRs the plots that removed the linear signal from the data had val set performance for the linear models that was consistently worse than random. I hypothesized that that was because the signal removal was being run across the full dataset, so any linear signal useful for predicting the training set was necessarily bad at predicting the val set or it would have been removed.

To test the hypothesis I simulated data generated by two classes. One dataset contains 5000 features: 2500 features with a linear boundary between the classes (each feature is drawn from a Gaussian with mean 0 if one class or mean 6 if the other) and 2500 features with a nonlinear boundary (one class is a zero centered Gaussian, the other is either drawn from a mean 6 Gaussian or a mean -6 Gaussian).

The results show that the nonlinear models are able to learn the nonlinear signal after linear signal has been removed, while the linear models have the overfitting problem.

image

image

Additionally, to ensure that the signal the nonlinear models are learning isn't just induced by limma, I ran another test with solely linear features. After correction, all models fail to perform better than random

image

image

Finally, I also used tried the first experiment when removing signal from train and val sets separately, and got the following result which indicates there is a bug in the skl LR wrapper's evaluation that is inflating performance. Since there is no linear dividing line in the data, it is impossible for a linear model to perform that well regardless of leakage, etc

image

Changes:
- Added relevant logic to Snakefile
- Added dataset configs for simulated datasets
- Visualized results
- Added logic for handling the new datasets to `predict_tissues`
- Wrote a script to simulate toy data


saged/simulate_be_data.py
```diff
@@ -0,0 +1,69 @@
...
+                current_var.append(np.random.normal(0))
+        vars['nonlinear_{}'.format(i)] = current_var
+
+    rownames = ['sample_{}'.format(i) for i in range(N_SAMPLES)]
+    data = pd.DataFrame(vars, index=rownames)
+    data['label'] = labels
+
+    if args.debug:
+        hist = data['linear_94'].hist(bins=40)
+        plt.show()
```
jjc2718, line 65: Where do these plots get shown? Are you just running this script in your terminal?

notebook/analysis/benchmark_results.ipynb
```diff
@@ -6340,6 +6340,2017 @@
...
...bSE9YBdgQGB8yJEFjHoDEMp+sheF7X82Hd87pIQbV6lDdjrChCLTwgDDoHdG9z5VTSbWTOJit/RgIa1Vp1Mj3sgnKeglbEAidiS6AUPSmaRxAoxaBR3NmIqTmHKMVuoxlIWcC1gBXILtjsEEWUsjIxaHcaCnBa018sMBc1cM63ffK89ZSuV7gu40SoueTNpvBF1us+s06YcEJRK3pMEhZPW0clZeUj20WoFCOBIRbhdNNSiR27Qk0hRaUKAA0RauIoaYVRiqxWaJWM5qaJE6EGFHWyAujuQIOCWspefgMgrxTV1rEjdo66SOpOAhcKlSJsrZDred76uqxG1jOZDNlsdl3uKxbhWCNiOk5CVFEr9ucy5FZ5EA+CgHK5vC7bkhbtF/FisbhhC2g4ERzJwinddtBYzT4/Xa0Tx7ArE6CVwoowFsW4fJFy9sILlnSjzdjvG8WJMNWMyJuQskpGRBtak8tlCVcRrLrleW6sI19rkF/QBjSKLaXQUM5m1v3nbdQ+bzhHVG1wIH/2ccxbR2wU5dz6vLZfqtXu8ytiy9fnq5xqHUNGsgEHiwVyKbrS0Ra3Bq5mY0smX1g04dTzvEt3WYX1ZrNJs9lcl/s6FcWMWqE3Gdhh0gn1Wo2Dq1xOvVwuMzc3ty7bkhbGGDKZDJVKZUMmGk7FljPW4SRZoGokMF3Vf3k1+7zaiGg0YxouQrXq1xuxo6KFuWY6n64bvd83UsU5jjdjyjqZMBcAo1bINhsMrCJUdcvz3IkQxpbTVsgriBAQRWANc83Guv+8jdrnTRFqjRijz07urzmH1pq5aH1e2y/Vavf5sShGxY7dKplg2qjHnLRx6vqTN5xwLIppKE2pWCSuVdij9bqX7q3XQJvnpVH6TuG7RMUlBz2tkst+BZ1c0oy3esMuU3POcTy2aCCvoeocxyKbuvKkPqMpaMWcFeatMGuFXq1Ttyz8dtHuwrOws4VBiFPWnkcrxYjRFBTMWEcsMJLCcp4Q6DWKWZcE97oTYoHelD2OWIQ5K/SZpGVjj9EUtWbGpqssCeC0tUxZR8MlpVbTsetcLfA8b32k6xWuiwRKEXP2gB2TlF74X+jWmLcODeRaNbllo6mKo5HCsH6/Vvu2nE5WL70mE1BM4aXx7SCjFAZFoxXO41aZVRon/E47oSbS2faJWFI3oVEpxe7AMGiSSbIK2BuY1HWDUa
```
jjc2718: This result is definitely odd (for the sklearn logistic regression model). I saw something similar to this when I was batch correcting a subset of features, but it looks like your x-axis here is the sample count (not the number of features) so I wouldn't have expected to see the "bump" here.

I'm not sure why performance would increase, then decrease as more samples are added. If there's really some kind of linear signal left I'd expect to see it increase more uniformly. Weird!

jjc2718 (approved):
Code looks good! Interesting results. I also tried some simulations a while back and couldn't quite recreate what I was seeing in the real data, but maybe yours are more realistic.

saged/simulate_be_data.py
ben-heil: For the analyses the script is run in snakemake, but in development I used the debug flag to make sure the feature distributions look correct (i.e. bimodal for linear features and trimodal for nonlinear)

notebook/analysis/benchmark_results.ipynb
ben-heil: My guess is that it's two opposing factors: accuracy gets worse as samples increase because the models are able to overfit the training data more, and the other side is probably related to the issue with the SKL LR implementation

ben-heil:
> I also tried some simulations a while back and couldn't quite recreate what I was seeing in the real data, but maybe yours are more realistic.

Maybe, or the patterns I'm seeing in the real data are just simpler :)

PR #70, opened 2022-01-19
Gtex signal removed

ben-heil:
This PR is a pretty short one - it adds the Snakemake logic to run the pipeline on GTEx data with tissue signal removed.

Changes:
- Added snakefile rules for GTEx linear signal removal runs
- Visualized results


jjc2718 (approved):
Looks good!

Super small question - in this plot:

image

Do you think the decrease in performance for logistic regression is the same bug with the sklearn wrapper you saw with the simulated data? I guess that woiuld make sense since it doesn't happen for your pytorch implementation.

ben-heil:
> Do you think the decrease in performance for logistic regression is the same bug with the sklearn wrapper you saw with the simulated data? I guess that woiuld make sense since it doesn't happen for your pytorch implementation.

I think it's related; I believe the SKL version's ability to fit the data better gives it the ability to overfit the train set better leading to a faster decline

PR #71, opened 2022-01-20
Simulations without signal and more split limma runs

ben-heil:
Another quick PR, this one is based on a suggestion from Marylyn to try the pipeline with completely random data to kick the wheels.

Results:

Even on datasets without signal, you can still see the limma-induced overfitting in the training set (though this is rescued by be correcting the train and val sets separately).

I also haven't run into the weird increase in skl LR accuracy again, it seems to be an interaction between train-val splitting, be correction, nonlinear features, and skl. I no longer think it's what's driving the LR performance, though I have another lead

image
image

Changes:
- New rules in the snakefile 
- Add dataset config for signalless data
- Visualize results
- Add logic to `simulate_be_data` to allow uninformative features


jjc2718 (approved):
Looks good!

Maybe you want to add in the plot titles (for the new plots you added) something about these being random features? Right now you have a couple of plots titled "simulated data prediction" and it could be hard to remember at a glance which is which. Up to you, though.

ben-heil:
> Looks good!
> 
> Maybe you want to add in the plot titles (for the new plots you added) something about these being random features? Right now you have a couple of plots titled "simulated data prediction" and it could be hard to remember at a glance which is which. Up to you, though.

Sounds good, my current plan is to clean up this notebook once I figure out which results are going to end up in the manuscript

PR #72, opened 2022-01-24
Pytorch vs SKL Logistic Regression

ben-heil:
This PR makes pytorch and skl LR as close together as I can make them to see if I can explain the differences in their performance. Tl;dr, they're still different when using the same optimizer and regularization scheme

Relevant results:
1) Pytorch LBFGS > Adam (unsurprising), but it doesn't fully rescue the accuracy (surprisingish).
2) The optimizer seems to be the cause of the weird accuracy increase when doing signal removal in the train and val set separately
image (1)
image

Changes:
- Add tissue2 wildcard to fully disambiguate rules in snakefile
- Update skl LR config to avoid using l2 norm by default
- Add config file for training pytorch LR with LBFGS
- Update models.py to allow the closure-style training required by LBFGS
- Visualize results


saged/models.py
```diff
@@ -227,11 +228,18 @@ def __init__(self,
...
         """
-        self.model = sklearn.linear_model.LogisticRegression(random_state=seed,
-                                                             class_weight='balanced',
-                                                             penalty='l2',
-                                                             C=l2_penalty,
-                                                             solver=solver)
+        if penalty_type == 'none':
+            self.model = sklearn.linear_model.LogisticRegression(random_state=seed,
+                                                                 class_weight='balanced',
+                                                                 penalty='none',
```
jjc2718, line 234: Do you want to add an option for `solver` here? Looks like the option will only take effect if you pass a `penalty_type` other than `'none'`.

saged/models.py
```diff
@@ -1364,17 +1376,32 @@ def fit(self, dataset: LabeledDataset, run: neptune.Run = None) -> "PytorchSuper
...
+                        output = self.model(expression)
 
-                loss = self.loss_fn(output, labels)
+                        loss = self.loss_fn(output, labels)
 
-                loss.backward()
+                        loss.backward()
+                        return loss
 
-                if getattr(self, 'clip_grads', False):
```
jjc2718, line 1374: Is gradient clipping not necessary for L-BFGS? Or do you just not want to use it?

jjc2718 (approved):
Looks good! It does inspire confidence that using L-BFGS brings pytorch closer to sklearn, although it is interesting that it doesn't fully rescue the accuracy.

Maybe it's the batch updates (pytorch) vs. no batches (sklearn)? Or are there any other differences between the methods that you know of at this point?

ben-heil:
> Looks good! It does inspire confidence that using L-BFGS brings pytorch closer to sklearn, although it is interesting that it doesn't fully rescue the accuracy.

Yep!

> Maybe it's the batch updates (pytorch) vs. no batches (sklearn)? Or are there any other differences between the methods that you know of at this point?

Nope, LBFGS isn't really an online algorithm as far as I can tell, so I used the full training set in the train loop by setting `full_batch_training` to true in the config

There are no remaining differences that I know of other than the LBFGS implementation and any other implementation details I don't know about since I've only read some of the skl source code

PR #73, opened 2022-02-09
Clean up figures

ben-heil:
This PR makes the current set of figures to be used in the manuscript, and separates out the analyses that didn't make the cut. The old figures at the bottom will be deleted in a future PR once the manuscript has solidified more

Changes:
- Make both sex prediction analyses use `predict_tissues.py`
- Clean up figures
- Add signal removed recount3 multiclass plot


jjc2718 (approved):
Looks good to me!

Are you planning to assemble multi-panel figures elsewhere (e.g. Inkscape or something like that)? If so, it might be good to label the paper figures these plots are part of once you decide on the final figures.

ben-heil:
> Are you planning to assemble multi-panel figures elsewhere (e.g. Inkscape or something like that)? If so, it might be good to label the paper figures these plots are part of once you decide on the final figures.

I'll end up modifying this notebook to include the final versions of figures (and labeling them then is a good idea). I just wanted to have a PR that was a clean list of what I had to work with before I started deleting and relabeling things

PR #74, opened 2022-03-01
Figure Update / TCGA Results

ben-heil:
This PR does two things (because I forgot to make a PR after the first thing :sweat_smile: )

First, it visualizes the results of running a TCGA mutation prediction pipeline (which works, though the weirdness with signal removal that you saw is still there). 

Second, it cleans up the figures/fills in missing figures that should have been run awhile ago

The TCGA figures/code probably won't make the final manuscript, so you don't need to review them too carefully. The main changes are the new figures in benchmark_results (the sex-prediction signal removal ones), though you might also be interested in the TCGA figures to see how they compare to the results that you saw


notebook/converted/benchmark_results.py
```diff
@@ -483,179 +439,1126 @@
...
-plot += ggtitle('Sex prediction (sample split)')
+plot += ggtitle('Metadata sex prediction signal removed')
+plot += xlab('Train sample count')
+plot += ylab('Balanced Accuracy (Val Set)')
+plot.save('../../figures/sex_prediction_signal_removed.svg')
 plot
 
 
-# ### Study Level
+# ### Study level split signal removed
```
jjc2718, line 515: As a reminder, what is "split signal removed"? Is this where you apply the batch correction to the training and test sets independently?

Just trying to figure out why the nonlinear models perform worse for this as opposed to the standard "signal removed" experiment. Did you see this for tissue prediction too?

notebook/converted/benchmark_results.py
```diff
@@ -731,17 +1634,17 @@
...
             
     tissue_metrics = pd.concat([tissue_metrics, new_df])
     
 tissue_metrics['train_count'] = tissue_metrics['train sample count']
-tissue_metrics['supervised'] = tissue_metrics['supervised'].str.replace('pytorch_supervised', 'three_layer_net')
+tissue_metrics['Model'] = tissue_metrics['Model'].str.replace('pytorch_supervised', 'three_layer_net')
 
 
-# In[49]:
+# In[102]:
```
jjc2718, line 1647: Looks like there's an error in this cell (102) - I'm seeing `NameError: name 'supervised' is not defined`.

notebook/converted/benchmark_results.py
```diff
@@ -775,151 +1678,166 @@
...
+plot += geom_hline(yintercept=.5, linetype='dashed')
 plot += geom_smooth()
-plot += geom_point(alpha=.2)
-plot += ggtitle('All Tissue Prediction GTEx Data (Patient-level split)')
+#plot += geom_point(alpha=.2)
+plot += facet_wrap(['gene'])
+plot += ggtitle('TCGA mutation prediction')
+plot += xlab('Train sample count')
+plot += ylab('Balanced Accuracy (Val Set)')
+plot.save('../../figures/tcga.svg')
```
jjc2718, line 1747: Interesting! I'm surprised that the five-layer net does so much better in the case of EGFR/KRAS/SETD2. I've tried random forests for those genes and didn't see much of an improvement over elastic net, but there were a few differences (we looked at AUPR rather than balanced accuracy, we used some additional covariates, we did cancer type filtering which I'm not sure if you did, etc).

I'm planning to try a shallow-ish NN for our multi-omics problem, so maybe I'll try it for gene expression too and see if I get similar results to you here.

Your TP53 and IDH1 results are pretty similar to what I would have expected, IDH1 is an easy classification problem and TP53 seems to be easy up until a point.

jjc2718 (approved):

notebook/converted/benchmark_results.py
ben-heil: Right, it's where the the batch correction is run on the training and test sets separately.

This pattern only happens here, my guess is that it happens randomly due to the colinearity in the data. Like limma has multiple potential ways to remove linear signal, and if it does so differently in the two sets, then the results become noisy

notebook/converted/benchmark_results.py
ben-heil: Good catch, fixed it!

notebook/converted/benchmark_results.py
ben-heil: Huh interesting. I think our problems are a little bit different since I'm not splitting out cancer types, it may be an artifact of that. Alternatively it could be that a smooth decision boundary is at as important as nonlinearity which is why the random forest doesn't look great but a five layer network does (but in that case I'd expect the three layer network to also do very well).

PR #75, opened 2022-03-18
Fix missing dependencies

ben-heil:
In the course of rerunning the code and optimizing hyperparameters, there have been a few missing libraries/undocumented dependencies/etc that needed to be added. This PR addresses those, so it's scope is kind of sprawly

Changes:
- Removed rules from snakefile that are unused in final version of analysis
- Added input file dependencies missing from some of the later rules in the snakefile
- Added plotnine, svgutils, and optuna to the env file
- Converted main scripts to use local imports instead of requiring users to install the repo as a package
- Fixed a breaking change where recount3 couldn't download because NAs were missing in the dataset or something by using `plyr`'s version of rbind instead of the base R version
- Fixed the normalization script for recount to calculate TPM as stated rather than RPKM
- Avoid saving models for every fold and seed to prevent out of disk space errors


Snakefile
```diff
@@ -485,45 +424,14 @@ rule all_tissue_prediction_signal_removed:
...
-        "results/all-tissue-biobert.{supervised}_{seed}.tsv"
-    shell:
-        "python saged/predict_tissue.py {input.dataset_config} {input.supervised_model} "
-        "results/all-tissue-biobert.{wildcards.supervised}_{wildcards.seed}.tsv "
-        "--neptune_config neptune.yml "
-        "--seed {wildcards.seed} "
-        "--all_tissue "
-        "--biobert "
-        "--weighted_loss "
+        # "--disable_optuna "
```
jjc2718, line 427: Do you want to leave these commented-out lines in? (In other words, is there a case where you might want to run this script with optuna disabled, or will you always run it with optuna at this point?)

jjc2718: Or maybe you could add an argument to the whole workflow that determines whether or not to do hyperparameter optimization? (Not sure if this is a thing you can do in Snakemake but my guess is yes?)

saged/download_recount3.R
```diff
@@ -10,12 +10,6 @@
 # recount3 from being installed. I recommend running this script from a base r installation 
 # rather than a conda env
 
-#install.packages("BiocManager", repos = "https://cran.r-project.org", dependencies=TRUE)
```
jjc2718, line 13: Maybe you could leave these install commands in but wrap them in a check to detect if they're already installed - something like [this might work?](https://stackoverflow.com/a/19871106)

Not sure what the convention is for this in R, though - maybe it's fine to just assume users already have the packages installed.

saged/predict_tissue.py
```diff
@@ -428,7 +430,7 @@ def prep_tcga_data(args: argparse.Namespace) -> Tuple[datasets.RefineBioMixedDat
                 if args.weighted_loss:
                     loss_weights = utils.calculate_loss_weights(train_data)
                     supervised_config['loss_weights'] = loss_weights
-                if 'save_path' in supervised_config:
+                if 'save_path' in supervised_config and i == 0 and args.seed == 0:
```
jjc2718, line 434: Might be good to add a comment explaining what you're doing here - looks like you only want to save the model info for the first iteration?

jjc2718 (approved):
Looks good! A few small questions/comments.

Snakefile
ben-heil: I think I'll leave them in; there might be a time in the future where I need to run them without optuna. I'm currently planning to remove them some time down the line when I finish cleaning up the Snakefile

saged/download_recount3.R
ben-heil: This script was originally meant to be part of a standalone package,  which is why I included the libraries used. Since I never ended up making the package, I'm fine with leaving all the installation up to Renv

saged/predict_tissue.py
ben-heil: Good idea, will do

PR #76, opened 2022-05-09
Hyperparameter optimization results

ben-heil:
This PR visualizes the results from the hyperparameter optimization run and adds the Snakefile used on the cluster

Changes:
- Remove unused rules from Snakefile
- Add figures to repo
- Update benchmark_results.ipynb to visualize the hyperparameter optmized results and adjust some figures


slurm_snakefile
```diff
@@ -0,0 +1,685 @@
...
+    # Random seeds should be numbers
+    seed="\d+"
+
+rule all:
+    input:
+        "data/batch_sim_data.tsv",
+        "data/linear_batch_sim_data.tsv",
+        "data/no_signal_batch_sim_data.tsv",
+	# Recount Binary classification
+        expand("/scratch/summit/benheil@xsede.org/results/{tissues}.{supervised}_{seed}.tsv",
```
jjc2718, line 32: Could you set this directory `/scratch/summit/benheil@xsede.org/results/` in a variable at the top of your script or something like that, since it's used throughout the script? Or are you fairly confident it'll never need to be changed?

jjc2718 (approved):
Looks good! Just one small comment.

slurm_snakefile
ben-heil: I'm pretty sure it'll never need to be changed. I only included this version of the snakefile to show which rules I ran on the cluster; the other Snakefile is the one people should actually use if they're going to run the pipeline.

I need to make sure to document that fact when I do my PR cleaning up the repo, so thanks for pointing it out!

PR #78, opened 2022-09-20
Out-of-sample generalization experiments

ben-heil:
A reviewer requested that we look at whether there were any differences between the out-of-sample prediction performances of linear and nonlinear models. This PR adds code to make that possible, updates the pipeline, and visualizes the results. 

@jjc2718 you might find the results interesting! There are some surprising trends, like having higher accuracy in the target domain than the source domain when using a model trained on recount to to binary prediction of GTEx tissues. 

The big changes are the `transfer_model` script which runs the experiments and the new cells at the bottom of visualize_results.ipynb.


Snakefile
```diff
@@ -33,6 +37,28 @@ result_files = [
             seed=range(0,NUM_SEEDS),
             tissues=GTEX_TISSUE_STRING,
             ),
+    #TODO Recount Transfer Binary classification
```
jjc2718: I assume this is done now?
```suggestion
    # Recount Transfer Binary classification
```

src/normalize_recount_gtex_genes.py
```diff
@@ -0,0 +1,248 @@
+"""This script normalizes the data from recount3"""
```
jjc2718: Can you just briefly mention which parts of this script are different than `src/normalize_recount_data.py`? It looks like it's mostly the same code, but with a filtering step to keep genes that are in both recount and GTEx - is that right?

src/transfer_model.py
```diff
@@ -0,0 +1,458 @@
...
+        tissue_1 = args.tissue1.replace('_', ' ')
+        tissue_2 = args.tissue2.replace('_', ' ')
+        labels_to_keep = [tissue_1, tissue_2]
+
+    labeled_data.subset_samples_to_labels(labels_to_keep)
+
+    return all_data, labeled_data
+
+
+def prep_sim_data(args: argparse.Namespace) -> Tuple[datasets.RefineBioMixedDataset,
```
jjc2718: Looks like this function isn't being used anywhere (as far as I can tell)

src/transfer_model.py
```diff
@@ -0,0 +1,458 @@
...
+                        help='The first tissue to be predicted from the data',
+                        default='Blood', choices=AVAILABLE_TISSUES)
+    parser.add_argument('--tissue2',
+                        help='The second tissue to be predicted from the data',
+                        default='Breast', choices=AVAILABLE_TISSUES)
+    parser.add_argument('--all_tissue', help='Predict all common tissues in the dataset',
+                        default=False, action='store_true')
+    # Recount only args
+    parser.add_argument('--biobert', help='Add biobert embeddings as features the model can use',
+                        default=False, action='store_true')
```
jjc2718: Are you planning to use this argument anywhere, or was it copy/pasted from an older script?

jjc2718 (approved):
Looks good! Definitely some interesting results here. Just a few small questions/comments.

src/transfer_model.py
ben-heil: Right, it's a leftover from the original script where I removed the part that called it but not the function itself. I'll go clean it (and the next comment) up

ben-heil:
<p>No reason in particular, just that I already had most of the infrastructure to run these analyses and didn't particularly want to build more for and extra plot in the response to reviewers. I agree that it would be a good experiment to run if this was ending up in the paper though</p>

---

 <a href='https://app.reviewnb.com/greenelab/linear_signal/pull/78/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ben-heil:
<p>True, I was mostly surprised when the recount -&gt; GTEx task had some problems where the out-of-sample GTEx domain had a higher accuracy than the training data on some problems. I agree that it's most likely just a data cleanness issue though</p><p> </p><p>The source measurements are from the held-out CV fold in the training data (so it's val accuracy, not train accuracy)</p>

---

 <a href='https://app.reviewnb.com/greenelab/linear_signal/pull/78/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

ben-heil:
<p>I definitely think it would be interesting! I don't think there are enough blood vessel samples in one of the datasets (GTEx iirc), but when I was looking at the distribution of errors in the multiclass task Blood and Blood Vessel samples were frequently being mistaken for each other. There may be some transfer learning that is doable there</p>

---

 <a href='https://app.reviewnb.com/greenelab/linear_signal/pull/78/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

